Issue #8, opened 2022-06-24
Configurable layouts for pages


rufuspollock:
Allow for (at least) 2 different layouts: a) "content/documentation" (classic markdown style) b) "landing" (or just "unstyled") which is pretty much a raw page with navbar and footer ie. unstyled

## Acceptance

* [x] can set `layout: {name}` in frontmatter of a markdown page and uses that layout from `components` (ignorecase is supported e.g. layout: abc` matches `layouts/Abc.jsx` **✅2022-06-28 in #15. Two layouts. @khalilcodes one question is whether case insensitive (not a big deal if not)**
* [x] One default one (maybe configurable??) that is used if not specified. Suggest docs is the default. **✅2022-06-28 docs is the the default**
* [x] Two layouts: `unstyled` and `docs` **✅2022-06-28 working**
* [x] Tests (at least "demo tests") **very basic tests. need something better if we can**

## Tasks

...

rufuspollock:
@khalilcodes i think all done - but leave you to confirm and close.

rufuspollock:
FIXED. All working. See `templates/default/layouts`

PR #15, opened 2022-06-28
[templates/default]: layouts for pages

khalilcodes:
## Classic docs and unstyled layouts for pages

Tasks from #8 
***

This PR adds support for pages to have different layouts if specified in frontmatter. There are two layouts `unstyled` and `docs`

### Tasks

- [x] Frontmatter field `layout: ...` (defaults to 'docs' if not specified)
- [x] Layouts folder with 'unstyled' and 'docs' - `layouts/unstyled.js` & `layouts/docs.js`
- [x] demo pages

### Previews

- Two demo pages in markdown `content/demo`
   - unstyled at [/demo](https://deploy-preview-15--spectacular-dragon-c1015c.netlify.app/demo)
   - docs at [/demo/docs](https://deploy-preview-15--spectacular-dragon-c1015c.netlify.app/demo/docs)

### DIFF
--- a/templates/default/tests/layouts.spec.js
+++ b/templates/default/tests/layouts.spec.js
@@ -1,0 +1,9 @@
+const { test, expect } = require("@playwright/test")
+
+test.describe('Layout tests', () => {
+    
+    // Test if the default page layout comes from layouts/docs.js
+    test('Default layout', async ({ page }) => {
+        
+    })
+})
--- a/templates/default/tests/pages.spec.js
+++ b/templates/default/tests/pages.spec.js
@@ -24,13 +24,13 @@ test.describe("Pages", () => {
   // })
 
   // Test for nested index.md routes
-  test('Nested index routes [test/index.md]', async ({ page, baseURL }) => {
+  test('Nested index routes [demo/index.md]', async ({ page, baseURL }) => {
     const Page = new MarkdownPage(page);
-    await Page.goto("/test");
+    await Page.goto("/demo");
     await Page.getData(page);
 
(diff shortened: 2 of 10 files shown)


templates/default/tests/layouts.spec.js
```diff
@@ -0,0 +1,9 @@
+const { test, expect } = require("@playwright/test")
+
+test.describe('Layout tests', () => {
+    
+    // Test if the default page layout comes from layouts/docs.js
+    test('Default layout', async ({ page }) => {
+
```
rufuspollock, line 7: this seems to be an empty test. generally not worth committing empty tests (?)

templates/default/tests/pages.spec.js
```diff
@@ -24,13 +24,13 @@ test.describe("Pages", () => {
   // })
 
   // Test for nested index.md routes
-  test('Nested index routes [test/index.md]', async ({ page, baseURL }) => {
+  test('Nested index routes [demo/index.md]', async ({ page, baseURL }) => {
     const Page = new MarkdownPage(page);
-    await Page.goto("/test");
```
rufuspollock, line 29: I would have kept the test page as we may use it for other things. IMO demos should be for demoing (ie. something we'll show users) and test pages for testing. That said, for now we could merge if that is useful as often doing the same thing. But in that case i'd stick to test i think.

Minor point!

rufuspollock (approved):

rufuspollock:
@khalilcodes again just want to acknowledge the absolutely perfect PR description - all the info i need, convenient links etc.

Issue #20, opened 2022-07-05
MDX try-out and tutorial with components and data


rufuspollock:
To make richer markdown pages we want MDX and be able to use components and data in the page e.g.

```
import MyComponent from '@components/MyComponent'

const orgs = [ 'abc' ]

# Hello world
```

## Acceptance

* ~[x] A demo page working using a component and data that is passed in e.g. `<MyCoolComponent value={props.mydata} />`~ **2022-08-11 (olayway:) we're no longer using demo pages**
* [X] Document these in a new page on flowershow.app e.g. `docs/howto-use-components-in-your-markdown` **✅ 2022-07-07 see https://flowershow.app/docs/howto-use-components-in-your-markdown** | **🚧 2022-08-11 (olayway:) 80% done, will include some context about MDX and how to import custom components/external data but also how to define them locally in a markdown file** | _**2022-08-15 done in PR #89**_
  - [X] How do we use components in the mdx file **🚧 2022-08-11 (olayway:) 80% done** | _**2022-08-15 done in PR #89**_
  - [X] How do we pass data into the mdx file and use in components **have to add the data in `pages/[[slug]].js`** | **🚧 2022-08-11 (olayway:) 80% done + no, you don't have to add the data in `pages/[[slug]].js` - you can import it from external file or define it directly in markdown** | _**2022-08-15 done in PR #89**_
- [X] Research if we really can import components and hero - see [Rufus comment below] **✅ 2022-08-22 (olayway:) yes, we can**(https://github.com/flowershow/flowershow/issues/20#issuecomment-1180376744)
  - [X] Report on whether it works or not by trying out a demo in our case (key will probably be getting import path right) **✅ 2022-08-22 (olayway:) currently you can see it working [here](https://flowershow.app/test/components-import)**

BONUS: find out if contentlayer supports direct imports in the MDX file e.g.

```
import MyComponent from '@components/MyComponent'

# Hello world
```
🚫 **tried but not working and could not find, but next-mdx-remote does so I may be missing something ?** **✅ 2022-08-22 (olayway:) yes, tested and works -> Contentlayer supports MDX and ESM `import`/`export` statements are part of MDX standard syntax, currently you can see it working [here](https://flowershow.app/test/components-import)**

## Tasks

- ~[ ] Demo page of component with props @ [/demo](https://flowershow.app/demo)~ **2022-08-11 (olayway:) we're no longer using demo pages**
[...]

PR #32, opened 2022-07-14
Per page props (data) import from json file

olayway:
As per discussion in issue #20, we don't want the user to change `[[...slug]].js`, since it complicates future template upgrades a lot. This approach seems to eliminate this problem:

## Changes
1.  Created `data/props.json` file whose purpose is to store all per-page data
  - key in `props.json` = `title` field value added to the document frontmatter
  - this file (empty) would be a part of the default template
2. `props.json` import in `[[...slug]].json`
  - extended `getStaticProps` to also find page-related data in the imported `props.json` file

## Requirements (from the user)
If a user wants to pass data to components in a given document, he/she needs to:
1. Add a `title` filed to a given document's front matter (I've used the already registered field but we could create another one for this purpose)
2. Add data he/she wants to pass to this document to `data/props.json` under the previously used `title`'s field value

## TODOs
- tweak `[[...slug]].js` code so that it's not necessary to keep an empty `props.json` file (i.e. removing it shouldn't result in an error)
- (maybe) parametrize path to this file so that the user can store this data under whatever name/directory (inside of project dir) he/she wants -> add this path to the `userConfig` file and use this import?

### DIFF
--- a/site/content/demo/index.md
+++ b/site/content/demo/index.md
@@ -1,5 +1,6 @@
 ---
 layout: unstyled
+title: demo
 ---
 
 `(layout/unstyled.js)`
--- a/templates/default/data/props.json
+++ b/templates/default/data/props.json
@@ -1,0 +1,18 @@
+{
+  "demo": {
+    "testData": [
+      {
+        "title": "First",
+        "value": 1
+      },
+      {
+        "title": "Second",
+        "value": 2
+      },
+      {
+        "title": "Third",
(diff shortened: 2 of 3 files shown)


olayway:
As discussed with @rufuspollock and @khalilcodes, we don't want to use additional JSON file (or js userConfig file) to specify data that should be passed to a given page. Instead, we want to enable the creation/manipulation of data objects directly in md files.

PR #52, opened 2022-07-19
renamed "userConfig" & disabled content path setting via env 

### LINKED ISSUE #7: Config can be overridden by user
When setting up my site I want to change configurable parameters so that I can change things without having to code

Also want to do this without getting into core code so that core remains update-able => separate config

## Acceptance

* [x] Can set a bootstrap config variable from environmental variables e.g. `FLOWERSHOW_CONTENT` **DONE. No need for dotenv as nextjs already supports .env / .env.local etc** **UPDATE 2022-07-19 we are going to remove this as not necessary for now**
* [x] Can set configuration from config file (TODO: location? is it .flowershow/config.js or simply config.js in content folder ...) **✅2022-07-19 we have gone (only feasible) path afaict of having a hard-coded config.js in `content/config.js`**

Bonus (separate)

* [x] Merge navLinks into siteConfig (no need for them to be separate) **✅2022-07-04 DONE**

## Tasks

* [x] Environment based configuration - maybe just for a bootstrap variable
* [x] File based configuration for the rest **✅2022-06-25 this turns out to be much more complex than thought so went for a very simple option of putting in a `userConfig.js` which is problematic in that user will have to get into the flowershow repo but is sort-of upgradeable**
* [x] Move `content/userConfig.js` to `content/config.js` **✅2022-07-19 PR #52**
* [x] Remove using env to set `content` folder in `siteConfg.js` and highlight in comments that it shouldn't be changed. **✅2022-07-19 PR #52**

## Notes

* Simplest to start with would be environment variables ...
* Better would be yaml or json config that user can have in the root of their project called e.g. flowershowconfig.yaml
* Easiest starting point might be working out how to override from env variables see e.g. https://github.com/motdotla/dotenv

On reflection env variables make sense for developers and therefore for app related stuff but users are unlikely to use them and instead will want a file.

That suggests that for the app config we really only need one config variable set by environment which is the location of the config file -- or the location of the content directory if the config file is always in a given location in that directory (e.g. `config.yml`

## File-based configuration - the saga 

Allowing for a user config file turns out to be much more complex thank I thought because it requires loading config from a dynamic location which isn't permitted by nextjs outside of getStaticProps (and hence not in e.g. `siteConfig.js` where we'd like to do it)

### Attempt 1

Try something like this:

```js
import * as path from 'path'
const configPath = path.resolve(content, 'config.js')   
const userConfig = await import(configPath)
const siteConfig = Object.assign({}, defaultConfig, userConfig)
```

As you get an error:

```
Module parse failed: Top-level-await is only supported in EcmaScript Modules
```

And if you try to fix that you get errors with webpack compilation e.g.

```
Critical dependency: the request of a dependency is an expression
```

### Attempt 2

What about if we were just loading json or yaml it may be easier ...

```js
import * as path from 'path'

import * as fs from 'fs'

const configPath = path.resolve(content, 'config.json')

const userConfig =  fs.existsSync(configPath) ?  JSON.load(fs.readFileSync(configPath)) : {}
```

But this runs into the issue that we can't use fs outside of getStaticProps etc ...

### So what to do?

Options

* [ ] Get config in e.g. `getStaticPaths` or ` _app.js` (or similar) and pass down to layouts
  * There's attraction to this as import from a given location is fragile with plugins anyway (e.g. can plugins easily find the file)
  * [ ] Do same but wrap in a `lib/config.js`
  * 🚩 have to be aware that there is no nice way in nextjs to load common data for whole app anywhere https://github.com/vercel/next.js/discussions/10909#discussioncomment-1266
    * The one way is to use `getInitialProps` though that disables "Automatic Static Generation" https://nextjs.org/docs/advanced-features/custom-app - however this is not a big issue for us i think as we are always using getStaticProps
* [ ] Quick and dirty: just have userConfig.js in config directory
  * But how does this get overridden? If it does not

olayway:
## Changes
- moves `content/userConfig.js` to `config.js`
- hardcodes `/content` as a path to markdown notes in `siteConfg.js`

Closes #7

### DIFF
--- a/templates/default/config/siteConfig.js
+++ b/templates/default/config/siteConfig.js
@@ -9,7 +9,10 @@ const defaultConfig = {
   // Google analytics key e.g. G-XXXX
   analytics: '',
   // content source directory for markdown files
-  content: process.env.FLOWERSHOW_CONTENT || 'content',
+  // DO NOT CHANGE THIS VALUE
+  // if you have your notes in another (external) directory,
+  // /content dir should be a symlink to that directory
+  content: 'content',
   // Theme
   navLinks: [
     { href: '/about', name: 'About' },
--- a/templates/default/config/siteConfig.js
+++ b/templates/default/config/siteConfig.js
@@ -31,7 +31,7 @@ const defaultConfig = {
 //  }
 }
 
-import userConfig from '../content/userConfig.js'
+import userConfig from '../content/config.js'
 
 const siteConfig = Object.assign({}, defaultConfig, userConfig)
 
--- a/site/content/config.js
+++ b/site/content/config.js
@@ -1,4 +1,4 @@
-const userConfig = {
+const config = {
   title: 'Flowershow',
   // Google analytics key e.g. G-XXXX
   analytics: 'G-RQWLTRWBS2',
@@ -10,4 +10,4 @@ const userConfig = {
   ],
 }
 
-export default userConfig
+export default config


templates/default/config/siteConfig.js
```diff
@@ -9,7 +9,10 @@ const defaultConfig = {
   // Google analytics key e.g. G-XXXX
   analytics: '',
   // content source directory for markdown files
-  content: process.env.FLOWERSHOW_CONTENT || 'content',
```
rufuspollock, line 12: if we remove content entirely won't things break?

rufuspollock (commented):
One significant item i've flagged.

templates/default/config/siteConfig.js
olayway: @rufuspollock, it's not really removed, this line has been changed. See lines below.

rufuspollock:
@olayway i can't merge now as has conflicts. Can you fix and then we merge?

Issue #26, opened 2022-07-12
Assets folder support


rufuspollock:
When I create stuff in `content/assets` (hard-code for now) i want that to show up on the site at `/assets` so my links work in my markdown

FUTURE: assets configurable and set up as part of scaffolding

## Acceptance

* [x] assets folder in flowershow.app works **✅2022-07-21 this is now working.**
* [x] rough docs for this (for now can explain you need to manually symlink and in future you do X) **✅2022-07-22 see https://flowershow.app/docs/essentials/assets**

## Tasks

* [x] Symlink in this repo
* [x] Add docs

rufuspollock:
@khalilcodes i think this is mostly done and we just need a docs page called e.g. `docs/assets` that explains that you put static files in a folder named assets ...

khalilcodes:
@rufuspollock added a rough doc in https://flowershow.app/docs/essentials/assets for symlink and using the assets folder.

rufuspollock:
I think we want this at simple `/docs/assets` and a brief explanation of what assets folder is for (e.g. static files). in obsidian/user content this would obsidian attachments. https://help.obsidian.md/How+to/Manage+attachments

olayway:
docs related to this to be moved under `/docs` folder

PR #58, opened 2022-07-21
[site/assets][s]: Assets folder support

khalilcodes:
## Create assets folder with symlink and add images
Related to tasks in #26 
***

This PR adds an assets folder to site and symlinks to public folder in `templates/default/public` so that image paths work in markdown.

### Tasks

- [x] Create assets folder in ~~`site/assets`~~ `site/content/assets`
- [x] Symlink the assets folder to `public/assets`
- [x] Move images to assets folder
- [x] update image links in for Hero and Logo


rufuspollock:
@khalilcodes this should have used the existing content assets folder not site assets folders i think.

khalilcodes:
@rufuspollock right I see that explained in the issue now. Will update that. 

Just to confirm do we still want it symlinked inside `templates/default/public` ?

rufuspollock:
@khalilcodes yes - remember the assets folder woudl get symlinked into public/assets (not directly into `public`)

PR #74, opened 2022-07-28
Section with currently supported and under development features

### LINKED ISSUE #65: Features / Roadmap section on frontpage
Add a features list / roadmap (i.e. features we haven't yet done) to the front page.

Can start by copying what we have on https://flowershow.app/docs/roadmap into frontpage and adding feature we have done with links to the docs

Do the basic first ie. use markdown and just headings etc

Later we can do another iteration to make it fancy

## Acceptance

**✅ 2022-07-28 all done in PR #74**

* [ ] Features and Roadmap section ✅
* [ ] Some items in it (don't have to list everything) ✅
  * [ ]  links to docs pages where they exist ✅
  * [ ] use emojis for under construction / planned / done ✅

## Tasks

* [x] Add section
* [x] Add 3 existing features
* [x] Add 3 planned features
* [ ] Review
* [ ] Then do more

olayway:
Closes #65

### DIFF
--- a/site/content/index.md
+++ b/site/content/index.md
@@ -119,24 +119,22 @@ layout: unstyled
   <div className="prose dark:prose-invert mx-auto max-w-2xl px-4 lg:max-w-6xl lg:px-8 xl:px-12">
     <h2 className="text-center">Features</h2>
 
-
 Here are some of the cool features that are currently supported by Flowershow, and some that we're intensively working on.
 
-### ✅/🚧 Markdown
+### ✅ Markdown
+
 Flowershow supports **CommonMark** and **GitHub Flavoured Markdown (GFM)** syntax, e.g. code blocks, blockquotes, lists, tasks lists and tables.
 
 [[syntax|Learn more ➡]]
 
-### ✅/🚧  Obsidian extensions to Markdown
-Flowershow has been designed with Obsidian users in mind, and so it aims to support the whole suite of Obsidian syntax extensions, like:
-- internal links ✅,
-- callouts 🚧,
-- math 🚧,
-- or Mermaid diagrams 🚧,
+### ✅  Obsidian wiki links
+
+Flowershow has been designed with Obsidian users in mind and so, currently, it supports internal links to other pages. Links to headings, blocks, and transclusions will also be supported in the future.
 
 [[syntax|Learn more ➡]]
 
 ### ✅ Custom page layouts and components
+
 You can customize and create new layouts for your pages.
 
 [[layouts|Learn more ➡]]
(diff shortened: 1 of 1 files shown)


site/content/index.md
```diff
@@ -119,16 +119,56 @@ layout: unstyled
...
-    - CommonMark, GitHub Flavoured Markdown and Obsidian extensions support (e.g. wikilinks) [[syntax|➡ learn more]]
-    - MDX support for custom React components additions [[custom-components|➡ learn more]]
-    - Customizable layouts of your pages [[layouts|➡ learn more]]
-    - Tailwind support for easy styling in-markdown HTML and custom components [[tailwind|➡ learn more]]
-    - Google analytics support [[analytics|➡ learn more]]
-    - Basic site configurations using single JSON file [[basic-config|➡ learn more]]
+Here are some of the cool features that are currently supported by Flowershow, and some that we're intensively working on.
 
-    👷 There is a lot of stuff we're still working on.
+### ✅/🚧 Markdown
```
rufuspollock, line 125: I'd aim to have either ✅ or 🚧 rather than both. Either that means splitting sections or going for one or other as appropriate.

On this one i'd say it was ✅ - ok there may be a few things but that isn't the end of the world.

site/content/index.md
```diff
@@ -119,16 +119,56 @@ layout: unstyled
...
+Flowershow supports **CommonMark** and **GitHub Flavoured Markdown (GFM)** syntax, e.g. code blocks, blockquotes, lists, tasks lists and tables.
+
+[[syntax|Learn more ➡]]
```
rufuspollock, line 128: I think we want to have the option in obsidian links to have full links otherwise these weren't render correctly e.g. i think this should be `[[docs/syntax| Learn more ...]]`

site/content/index.md
```diff
@@ -119,16 +119,56 @@ layout: unstyled
...
+
+### ✅/🚧  Obsidian extensions to Markdown
+Flowershow has been designed with Obsidian users in mind, and so it aims to support the whole suite of Obsidian syntax extensions, like:
+- internal links ✅,
+- callouts 🚧,
+- math 🚧,
```
rufuspollock, line 134: I'd say Math and Mermaid aren't obsidian extensions but rather complete features in their own right with their own section. I'd also say that for callouts in fact.

I'd add as obsidian syntax the transclusion stuff that we don't fully support yet i.e. `![[...]]` type syntax.

So i'd call this section Obsidian wiki links and put it as ✅ or 🚧 depending on how important transclusion is.

site/content/index.md
```diff
@@ -119,16 +119,56 @@ layout: unstyled
...
+- or Mermaid diagrams 🚧,
+
+[[syntax|Learn more ➡]]
+
+### ✅ Custom page layouts and components
+You can customize and create new layouts for your pages.
```
rufuspollock, line 140: Can we have a newline after headings. Minor markdown syntax preference.

rufuspollock (changes requested):
Some minor-ish suggestions.

site/content/index.md
olayway: I think our `remark-wiki-link-plus` is smart enough to handle this.
Basically Obsidian will automatically create a link in form `[[syntax]]` if there is only one page called `syntax`, no matter in what folder it's located. It would be a bummer for Flowershow to work a user would have to manually correct all links to include the full path.
In this case for example `[[syntax|Learn more ➡]]` is properly converted to `<a class="internal" href="/docs/syntax">Learn more ➡</a>`, without me having to explicitly point at the `/docs` dir.

site/content/index.md
olayway: I've rearranged it into: 

### ✅ Obsidian wiki links
and
### 🚧 Excalidraw, Mermaid an Mathjax support

site/content/index.md
rufuspollock: @olayway i don't think remark-wiki-link-plus handles that actually. There is a config option in obsidian setting to have it do relative paths from the base of the wiki.

This is actually something we probably need to note in the docs in our wiki links section that people must use absolute path or relative path for Flowershow published links to work.

image

site/content/index.md
rufuspollock: OK. Though i'd not that each of those is a feature in itself! 😄

rufuspollock:
@olayway this all looks good now. I'm assuming you still want to take care of merging?

site/content/index.md
olayway: I wanted to keep it concise on the front page so I've listed them under one heading, but I can split it if you wish :slightly_smiling_face:

site/content/index.md
olayway: Hmm, it is strange that it works locally in this case, but ok...
I'll add a short note to the docs.

rufuspollock (approved):

PR #75, opened 2022-07-28
`docs/syntax` page adjustments

olayway:
## Changes
- syntax elements in the `docs/syntax` page split into three sections: CommonMark, GFM and Obsidian extensions (based on CommonMark and GFM specs, and Obsidian docs)
- a few more syntax elements added
- added "callouts" at the end of each section pointing to CommonMark and GFM specs, and Obsidian docs at the end of corresponding sections (for users to learn more about each syntax and so that we don't need to write down everything that's already there)

### DIFF
--- a/site/content/docs/guides.md
+++ b/site/content/docs/guides.md
@@ -2,7 +2,7 @@
 
 Here are our basics guides, that will help you make the most of your Flowershow website:
 1. [[basic-config|Basic configurations with `config.js`]]
-2. [[syntax|Supported markdown and Obsidian syntax]]
+2. [[syntax|Markdown syntax support]]
 3. [[tailwind|Styling with Tailwind]]
 4. [[custom-components|Creating custom components]]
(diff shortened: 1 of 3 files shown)


rufuspollock:
I'd propose callouts as their own section and i'd have wiki-links (obsidian stuff) as their own section e.g. `docs/wiki-links`. NB: these are my opinions not a strong "we must do this" so if you want to stick with this setup that's ok with me 😄

olayway:
> I'd propose callouts as their own section and i'd have wiki-links (obsidian stuff) as their own section e.g. `docs/wiki-links`. NB: these are my opinions not a strong "we must do this" so if you want to stick with this setup that's ok with me smile

I think you're right, as neither callouts nor math nor mermaid are really Obsidian-specific features. I've created a section called "Other extensions" where I've put everything that doesn't belong in CommonMark or GFM for now, but maybe we could use some other, better name for this section?

PR #77, opened 2022-08-03
Math syntax support

### LINKED ISSUE #64: Math syntax support
Math support is very useful. We'll follow Obsidian who use mathjax (https://help.obsidian.md/How+to/Format+your+notes#Math).

Here's a guide to supporting this in markdown/mdx: https://mdxjs.com/guides/math/

Essence is to use two plugins https://github.com/remarkjs/remark-math/tree/main/packages/remark-math and https://github.com/remarkjs/remark-math/tree/main/packages/rehype-mathjax

## Acceptance

* [ ] Inline math like `$\sqrt{a^2 + b^2}$` works
* [ ] Block math works like the below
* [ ] Documentation on e.g. `/docs/math`

```
$$\begin{vmatrix}a & b\\
c & d
\end{vmatrix}=ad-bc$$
```

## Tasks

* [ ] Add the two plugins

khalilcodes:
## Add math syntax support with documentation
Resolves #64 
***
This PR adds support for using math syntax as in obsidian using the following remark and rehype plugins
- https://github.com/remarkjs/remark-math/tree/main/packages/remark-math
- https://github.com/remarkjs/remark-math/tree/main/packages/rehype-mathjax

### Tasks

- [x] add both remark and rehype plugins in contentlayer
- [x] ~~add custom Paragraph component to support some breaking styles with math syntax~~
- [x] add math inline styles in `styles/global.css`
- [x] create docs at [/docs/math](https://deploy-preview-77--spectacular-dragon-c1015c.netlify.app/docs/math)

### Notes

Block math only transforms when `$$` signs are above and below the syntax. for eg.
```markdown
$$
\begin{vmatrix}a & b\\
c & d
\end{vmatrix}=ad-bc
$$
```

### DIFF
--- a/site/content/docs/math.md
+++ b/site/content/docs/math.md
@@ -1,0 +1,42 @@
+# Math
+
+Flowershow supports math syntax where you can use LaTeX notations to denote formulas. It is using the  [MathJax](https://docs.mathjax.org/en/latest/basic/mathjax.html) package and displays them as svg.
+
+## How to use
+
+### Inline math
+
+* Use between `$`
+
+For example if your markdown file has the following
+
+```markdown
+$\sqrt{a^2 + b^2}$
+```
+will be transformed as
+
+$\sqrt{a^2 + b^2}$
+
+### Block math
+* Use between double `$$`
+
+For example if your markdown has the following
+
+```markdown
+$$
+\begin{vmatrix}a & b\\
+c & d
+\end{vmatrix}=ad-bc
+$$
+```
+
+will be tranformed as
+
+$$
+\begin{vmatrix}a & b\\
+c & d
+\end{vmatrix}=ad-bc
+$$
+
+
+
--- a/site/content/docs/math.md
+++ b/site/content/docs/math.md
@@ -1,6 +1,6 @@
 # Math
 
-Flowershow supports math syntax where you can use LaTeX notations to denote formulas. It is using the  [MathJax](https://docs.mathjax.org/en/latest/basic/mathjax.html) package and displays them as svg.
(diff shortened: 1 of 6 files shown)


site/content/docs/math.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+will be tranformed as
+
+$$
+\begin{vmatrix}a & b\\
+c & d
+\end{vmatrix}=ad-bc
+$$
+
```
rufuspollock, line 40: Trailing blank space. Maybe also add a short section for Developers e.g. `## Developers` that says that this uses mathjax or similar.

templates/default/components/MDX.js
```diff
@@ -2,9 +2,18 @@ import Head from 'next/head'
 import DemoComponent from './Demo'
 import { Pre } from './Pre'
 
+const Paragraph = (props) => {
```
rufuspollock, line 5: Why is this necessary? Looking at https://github.com/remarkjs/remark-math it seemed we could just use the remark plugins.

I'm guessing this is styling related. If so some comment here would have been recommended to explain this briefly.

Also if we used katex would this still be necessary?

And could we use traditional "css" for this? 

I ask as this seems a bit hacky and impinges on core code.

rufuspollock (commented):
Have one substantive question - otherwise looks good.

templates/default/components/MDX.js
khalilcodes: Yes, this was style related. Moved to css and removed the component.

rufuspollock:
@khalilcodes LGTM 👏 . Good work!

Issue #57, opened 2022-07-20
SEO support


rufuspollock:
When creating a site I want to set description, image, title set on page (and have default values) so that have good SEO 

There are really two parts to this:

- Site-wide SEO: common and/or default SEO values for the whole site (and especially the front page)
- Per-page SEO: SEO per page that overrides default SEO values. These would derive from frontmatter and/or be calculated e.g. title from markdown first heading etc (nb: any work to calculate title etc is out of scope for this issue - this issue can assume those values are available)

**Priority right now is the first**

We can follow the work we did in web3 for this e.g. https://github.com/life-itself/web3/issues/114

## Acceptance

* [x] Can configure default SEO - full stack of next-seo options in config and these get used on every page (can get overriden) **✅2022-08-16 see https://flowershow.app/docs/seo and https://socialsharepreview.com/?url=https://flowershow.app**
  * [x] This SEO shows up on front page **✅2022-08-16 see https://socialsharepreview.com/?url=https://flowershow.app**
  * [x] SEO shows up on every other page where not explicitly overridden **✅2022-08-16 this works**
  * [x] Docs page explaining how to set and referencing options in next-seo **✅2022-08-16 https://flowershow.app/docs/seo**
* [x] Per page SEO values for title, description, image etc which come from frontmatter or elsewhere **🚧2022-08-16 done for title, description but not image yet. An example can be see on https://flowershow.app/docs/seo where title is different**
  <img width="500" alt="image" src="https://user-images.githubusercontent.com/180658/184887045-8ca0ad46-03e6-494a-b580-c7c65b4f8a40.png">

Proposed Default Next SEO (note other SEO values would get set from other config values e.g. title and description)

```
nextSeo: {
    openGraph: {
      images: [
        {
          url: "https://web3.lifeitself.us/img/home-page-screenshot-social-preview.png",
          alt: "",
          width: 1200,
          height: 627,
          type: "image/jpg",
        },
      ],
    },
    twitter: {
      handle: "@forlifeitself",
      site: "web3.lifeitself.us",
      cardType: "summary_large_image",
    },
  },
```

## Tasks

* [x] Create docs
* [x] ...

rufuspollock:
This is high priority as we are about to start sharing on social media (at least for front page to have it).

rufuspollock:
@khalilcodes to add image support as last item.

olayway:
@rufuspollock, @khalilcodes I think this issue can be closed now?

[...]

PR #79, opened 2022-08-08
SEO support

khalilcodes:
## Add seo support to frontpage - title, description and image (social)
Tasks related to #57
***

Merging this PR will add support for image, title and description while sharing on social sites (twitter) with default configs which can be overriden by if user config values are present.

### Tasks
- [x] Add screen shot of frontpage to content's assets
- [x] Add twitter and social image url fields to user config `site/content/config`
- [x] Add default seo in `config/siteConfig` if user config is empty
- [x] Add NextSeo with frontmatter and nextSeo values from above


site/content/config.js
```diff
@@ -7,6 +7,11 @@ const config = {
     { href: '/docs/roadmap', name: 'Roadmap' },
     { href: '/docs/guides', name: 'Guides' },
   ],
+  socialImageUrl: '',
```
rufuspollock, line 10: Why are we adding blank values here? Surely these should be left out and we have defaults in siteConfig.js?

templates/default/config/siteConfig.js
```diff
@@ -21,23 +23,28 @@ const defaultConfig = {
...
-//      ]
-//    }
-//  }
+  nextSeo: {
+    openGraph: {
+      images: [
+        {
+          url: userConfig.socialImageUrl 
+            ? userConfig.socialImageUrl 
+            : 'https://flowershow.app/assets/images/frontpage-screenshot.jpg',
```
rufuspollock, line 32: I don't think we should default this like this for all users. This should be empty by default.

templates/default/config/siteConfig.js
```diff
@@ -21,23 +23,28 @@ const defaultConfig = {
...
+          alt: userConfig.title ? userConfig.title : '',
+          width: 1200,
+          height: 627,
+          type: 'image/jpg',
```
rufuspollock, line 36: Again defaulting like this could lead to issues: what happens if png?

templates/default/config/siteConfig.js
```diff
@@ -21,23 +23,28 @@ const defaultConfig = {
...
+        }
+      ]
+    },
+    twitter: {
+      handle: userConfig.twitter?.handle ? userConfig.twitter.handle : "@flowershow",
+      site: userConfig.twitter?.site ? userConfig.twitter.site : "https://flowershow.app",
```
rufuspollock, line 42: Again, let's not default to flowershow here etc.

templates/default/config/siteConfig.js
```diff
@@ -21,23 +23,28 @@ const defaultConfig = {
...
-//          url: 'https://image.url/...',
-//          alt: '',
-//          width: 1200,
-//          height: 627,
-//          type: 'image/png',
-//        }
-//      ]
-//    }
-//  }
+  nextSeo: {
```
rufuspollock, line 26: I'm not sure about this approach - i think we want to simply reproduce nextSeo stuff into e.g. userConfig rather than try to simplify it. This would means docs page for this would show what would be a template for what to copy into `config.js` and explain where it comes from.

Why? Because a) users can't access full nextSeo stuff b) we have to maintain the interface between our config and nextSeo.

rufuspollock (changes requested):
I think this wants to be simpler 😄

i.e. just have nextSeo stuff in `config.js` and as there is almost nothing to default here there's almost no point having stuff in `siteConfig.js` (the only possible item is setting the twitter care image).

Also missing here which is *crucial* are the docs - i'd suggest starting here with docs driven development where you wrote the docs first and put that in a PR and i reviewed that and then we implemented the code to support that.

khalilcodes:
@rufuspollock I can see with the last commits that per page SEO for title and image works (coming from frontmatter if specified) and defaults are set in user config. Are we looking to add per page image in frontmatter or are we just going with the default one for now ?

templates/default/config/siteConfig.js
khalilcodes: moved seo defaults to user config now and no seo in siteConfig.

rufuspollock:
Now all looking good. We can merge. Remaining is per page image.

PR #87, opened 2022-08-11
Support transclusion links for images

khalilcodes:
## Adding support for transclusion links (images)

This PR adds the feature to support transclusion links for images as used in Obsidian. It parses `![[Image.png]]` as an image element ``

### Tasks
- [x] bump remark-wiki-link-plus to latest package version which is now v1.0.2
- [x] create test page in [/test/page-4](https://deploy-preview-87--spectacular-dragon-c1015c.netlify.app/test/page-4)

### Notes

Docs for this support is mentioned in [/docs/syntax](https://flowershow.app/docs/syntax)

### Example screenshot
Screen Shot 2022-08-11 at 4 33 25 PM

### DIFF
--- a/templates/default/package.json
+++ b/templates/default/package.json
@@ -27,7 +27,7 @@
     "remark-math": "^5.1.1",
     "remark-slug": "^7.0.0",
     "remark-toc": "^8.0.0",
-    "remark-wiki-link-plus": "^1.0.1"
+    "remark-wiki-link-plus": "^1.0.2"
   },
   "devDependencies": {
     "@playwright/test": "^1.22.2",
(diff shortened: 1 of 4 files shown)


rufuspollock:
@khalilcodes looks good. one suggestion is that rather than having a test page could this just be part of the wiki links docs here https://flowershow.app/docs/syntax (with a live example)?

templates/default/package.json
```diff
@@ -27,7 +27,7 @@
     "remark-math": "^5.1.1",
     "remark-slug": "^7.0.0",
     "remark-toc": "^8.0.0",
-    "remark-wiki-link-plus": "^1.0.1"
+    "remark-wiki-link-plus": "^1.0.2"
```
rufuspollock, line 30: v minor: would normally have bumped minor version number rather than ultra-minor as significant feature e.g. `1.1.0` rather than `1.0.2`

Issue #12, opened 2022-06-27
Mermaid support


rufuspollock:
When creating markdown pages I want to add mermaid-syntax diagrams so that I can display nicely rendered diagrams 

https://mermaid-js.github.io/mermaid/#/

## Acceptance

* [x] I can have have mermaid blocks and they render correctly **✅2022-08-16 see https://flowershow.app/docs/mermaid**

```
```mermaid
graph TD

a --> b
\```
```

## Tasks

* [x] research options e.g. existing remark plugins
* [x] Select one to try
* [x] Try it
* [x] Deploy
* [x] Document
* [x] ...

## Notes

We want a solution that renders client side (we do *not* want to render svg and insert)

Options

* https://github.com/temando/remark-mermaid **2022-07-26 this looks pretty promising for what we want**
  **2022-08-02** *summary: adding as a plugin in remark just throws error as seen in notes below*
* https://github.com/sjwall/mdx-mermaid - Plug and play Mermaid in MDX
  **2022-08-02** *summary:*
  - adding as a plugin in remark just outputs the component as a string.
  - Adding the component in MDX components client side ie. `Pre.js` is complicated as `rehype-prism-plus` converts the graph to span elements and `<Mermaid char={...} />` needs the graph as a string.
* https://github.com/remcohaszing/remark-mermaidjs - A remark plugin to render mermaid diagrams using puppeteer
   **2022-08-02** *summary: not considering as puppeteer is a large dependency even if it works without much configuration needed*

### remark-mermaid

 - throwing errors: `Expected usable value, not 'undefined'` Looks like it's trying to parse wrong values.
 
  ![Screen Shot 2022-07-26 at 6 00 07 PM](https://user-images.githubusercontent.com/42637597/181024708-2fa0d6ca-6652-494d-9a28-db358fbc8188.png)

### mdx-mermaid with remark-code-extra (UPDATE)

Using this remark plugin - https://github.com/s0/remark-code-extra - to replace `pre` tag elements created by `rehype-prism-plus`, I have managed to make `mdx-mermaid` work nicely with it. See screenshots below for result and modified code in `contentlayer.config.js`

*Brief summary of modified code:*

Transforming the nodes for `<Pre>` elements and removing the ones created by rehype-prism-plus ...

- Added the remark plugin in contentlayer with following configuration within it:
   - create a new pre tag element with just the string value as child and a new className `.code-mermaid`
   - remove the previous pre tag element created by rehype-prism-plus
[...]

PR #90, opened 2022-08-12
Mermaid support

khalilcodes:
## Add mermaid support to markdown
Tasks from #12 
***

This PR adds the feature to convert mermaid code blocks to diagrams using the following remark packages 
- https://github.com/sjwall/mdx-mermaid
- https://github.com/s0/remark-code-extra

### Tasks
- [x] Modify `<Pre>` element nodes using remark-code-extra plugin in contentlayer so that it works with the mdx-mermaid component
- [ ] Add some info to above in comments on why we are using this approach and link issue
- [x] Add Mermaid component in `Pre.js` 

### Notes

Docs for mermaid present in https://flowershow.app/docs/syntax

### Screenshot preview
Screen Shot 2022-08-12 at 2 47 01 PM


templates/default/components/Pre.js
```diff
@@ -20,6 +21,14 @@ export const Pre = ({ children }) => {
     }, 2000);
   };
 
+  if (props.className && props.className === "code-mermaid") {
```
rufuspollock, line 24: We really want some docs here as to why we are doing this and e.g. link to github issue.

I mean i can guess a bit but more info is useful.

templates/default/contentlayer.config.js
```diff
@@ -57,7 +59,45 @@ export default makeSource({
     remarkPlugins: [ 
       remarkGfm,
       remarkMath,
-      [ wikiLinkPlugin, { markdownFolder: siteConfig.content } ]
+      [wikiLinkPlugin, { markdownFolder: siteConfig.content }],
+      /** Using the code extra plugin from https://github.com/s0/remark-code-extra
```
rufuspollock, line 63: Good comment.

rufuspollock (approved):
Looks good. Well done on getting this working.

The one issue i notice (we can look at later) is that mermaid coloring does not work well with dark theme

image

PR #96, opened 2022-08-16
Add e2e tests in github action

abhishekgahlot:


templates/default/tests/wikilinks.spec.js
```diff
@@ -1,70 +0,0 @@
-const { test, expect } = require("@playwright/test")
```
rufuspollock, line 1: Why is this getting deleted?

rufuspollock:
LGTM.

PR #121, opened 2022-08-24
Section about MDX components in MDX docs page

### LINKED ISSUE #110: Add section about MDX components to the mdx doc page
Atm we have it mentioned only on one of the test pages. This should also be described in the docs page related to MDX.

olayway:
Closes #110

---

## Changes
- created a section about MDX components (in mdx docs page) and removed the corresponding test page


site/content/docs/mdx.md
```diff
@@ -299,3 +299,119 @@ You can also use comments in expressions with JavaScript's multiline comment syn
 ```md
 {/* A comment! */}
 ```
+
+## MDX components
```
rufuspollock, line 303: @olayway the one thing i'm not clear about reading this is **why** i would use mdx vs jsx extension for my components. i.e. when would i use "MDX" components vs standard components? is it when i want to use markdown? Would it be worth adding a very brief explanation of *why* you would use mdx components vs standard jsx ones?

rufuspollock:
@olayway see my comment above - would be happy to merge without any changes if either a) you don't think explanation needed or b) we do that later.

site/content/docs/mdx.md
olayway: You're right, I'll add a few sentences :)

rufuspollock:
@olayway LGTM 👏

PR #123, opened 2022-08-24
[SEO]: Seo fixes and default configs

khalilcodes:
### SEO titles in config
#57 
***

This PR adds the feature to set nextSeo title in config and fixes the default seo image overrides.

### Tasks
- [x] set title and description in user config file `config.js` for default seo
- [x] fix images overrides and url
- [x] remove nextSeo from layout

### Notes
- There is no `titleTemplate` here as suggested in #57 comments. If we are setting a `titleTemplate` in config, then it appears twice for pages without a frontmatter title. This is because the title is set in config already and is never undefined.
  Screen Shot 2022-08-24 at 3 44 24 PM

### DIFF
--- a/site/content/config.js
+++ b/site/content/config.js
@@ -1,6 +1,5 @@
 const config = {
   title: 'Flowershow',
-  description: 'Turn your markdown notes into an elegant website and tailor it to your needs. Flowershow is easy to use, fully-featured, Obsidian compatible and open-source.',
   analytics: 'G-RQWLTRWBS2',
   navLinks: [
     { href: '/#overview', name: 'Overview' },
@@ -10,7 +9,11 @@ const config = {
     { href: 'https://github.com/flowershow/flowershow/discussions', name: 'Forum' },
   ],
   nextSeo: {
+    title: 'Flowershow',
+    description: 'Turn your markdown notes into an elegant website and tailor it to your needs. Flowershow is easy to use, fully-featured, Obsidian compatible and open-source.',
+    canonical: 'https://previews-823b8d.netlify.live',
     openGraph: {
+      title: 'Flowershow',
       images: [
         {
           url: 'https://flowershow.app/assets/images/frontpage-screenshot.jpg',
--- a/templates/default/components/MDX.js
+++ b/templates/default/components/MDX.js
@@ -1,7 +1,7 @@
 import { NextSeo } from 'next-seo'
 import Head from 'next/head'
 import { Pre } from './Pre'
-import config from '../content/config'
+import siteConfig from '../config/siteConfig'
 
 const components = {
   Head,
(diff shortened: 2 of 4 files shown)


templates/default/components/MDX.js
```diff
@@ -15,24 +15,25 @@ const components = {
 
 export default function MdxPage({ children, ...rest }) {
   const { Component, frontMatter } = children;
+  const seoImageUrl = siteConfig.authorUrl + frontMatter.image
```
rufuspollock, line 18: What happens if frontMatter.image is an absolute url and not relative?

site/content/config.js
```diff
@@ -10,7 +9,11 @@ const config = {
     { href: 'https://github.com/flowershow/flowershow/discussions', name: 'Forum' },
   ],
   nextSeo: {
+    title: 'Flowershow',
+    description: 'Turn your markdown notes into an elegant website and tailor it to your needs. Flowershow is easy to use, fully-featured, Obsidian compatible and open-source.',
+    canonical: 'https://previews-823b8d.netlify.live',
```
rufuspollock, line 14: This seems an odd canonical url - surely it's https://flowershow.app/

templates/default/pages/_app.js
```diff
@@ -29,12 +29,7 @@ function MyApp({ Component, pageProps }) {
 
   return (
     <ThemeProvider attribute="class" defaultTheme="dark">
-      <DefaultSeo
-        titleTemplate={'%s | ' + siteConfig.title}
```
rufuspollock, line 33: I think there is now an issue as i don't see that we are getting right title e.g. if i go to https://deploy-preview-123--spectacular-dragon-c1015c.netlify.app/docs/roadmap i just get "Roadmap" in the title not "Roadmap | Flowershow"

rufuspollock (changes requested):
Some stuff still to resolve IMO.

site/content/config.js
khalilcodes: Yes! Forgot to change that as I was testing. Thanks for pointing out !!

templates/default/components/MDX.js
khalilcodes: I will add the check for this 👍

templates/default/pages/_app.js
khalilcodes: Adding a templateTitle will do this for the homepage or any other page that doesn't have a frontmatter title because the title is always defined in `config.js` nextSeo property.

Screen Shot 2022-08-24 at 3 44 24 PM

khalilcodes:
@rufuspollock @olayway new commits for above fixes pushed plus added some documentation about setting authorUrl in user config (in `siteConfig` the default is flowershow.app which is important to change for users).

PR #158, opened 2022-09-11
added edit link at bottom of pages

PhilippeduPreez:
- Edit Page link added to the bottom of pages
- Edit this page can be enabled / disabled (disabled by default) in config
- Edit this page can be enabled / disabled per page via frontmatter option

### DIFF
--- a/templates/default/config/siteConfig.js
+++ b/templates/default/config/siteConfig.js
@@ -3,6 +3,9 @@ import userConfig from '../content/config.js';
 const defaultConfig = {
   title: 'Flowershow',
   description: 'Publish your digital garden',
+  repoRoot: "https://github.com/flowershow/flowershow",
+  repoEditPath: "/edit/main/site/content/",
+  editPageShow: false,
   author: 'Flowershow',
   // logo image
   authorLogo: '/assets/images/logo.svg',
--- a/templates/default/contentlayer.config.js
+++ b/templates/default/contentlayer.config.js
@@ -16,6 +16,7 @@ const sharedFields = {
   description: { type: 'string' },
   image: { type: 'string' },
   layout: { type: 'string', default: 'docs' },
+  editLink: { type: 'boolean', default: false},
 };
 
 const computedFields = {
--- a/templates/default/layouts/docs.js
+++ b/templates/default/layouts/docs.js
@@ -1,5 +1,9 @@
+import siteConfig from '../config/siteConfig';
+
 export default function DocsLayout({ children, frontMatter }) {
-  const { title } = frontMatter;
+  const { title, editLink } = frontMatter;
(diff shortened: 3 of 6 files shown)


templates/default/config/siteConfig.js
```diff
@@ -3,6 +3,9 @@ import userConfig from '../content/config.js';
 const defaultConfig = {
   title: 'Flowershow',
   description: 'Publish your digital garden',
+  repoRoot: "https://github.com/flowershow/flowershow",
+  repoEditPath: "/edit/main/site/content/",
+  editPageShow: false,
```
olayway, line 8: Let's maybe call this `editLinkShow`

templates/default/contentlayer.config.js
```diff
@@ -16,6 +16,7 @@ const sharedFields = {
   description: { type: 'string' },
   image: { type: 'string' },
   layout: { type: 'string', default: 'docs' },
+  editLink: { type: 'boolean', default: false},
```
olayway, line 19: Having this default value set to false independently of the value in config, results in the need to manually change both values in order to add the link to docs pages (without the need to use frontmatter).

olayway: If this value comes from the config, the problem is solved.

templates/default/layouts/docs.js
```diff
@@ -1,12 +1,41 @@
...
+  // edit link will only appear if enabled in frontmatter and in siteConfig
+  const editUrl = siteConfig.repoRoot + siteConfig.repoEditPath + frontMatter._raw.sourceFilePath;
   return (
     <article className="docs prose dark:prose-invert prose-a:break-words mx-auto p-6">
       <header>
         <div className="mb-6">{title && <h1>{title}</h1>}</div>
       </header>
       <section>
         {children}
+        {editUrl && siteConfig.editPageShow && editLink && (
```
olayway, line 14: ...so, in this case if I switch `editPageShow` to `true` but in `configlayer.config` we have `editLink` still set to false, the button will not show up.

olayway (changes requested):
See my comments related to the on/off logic.

Additionally, after giving it a second thought I think we should move this button to `Layout` component, instead of having it in the `docs` layout. This is because at this point the edit button will only show up on the docs pages and in the future, we will probably also have other document types used, e.g. blog. But this is maybe for another time as I think we will need to do some tidying up in the Contentlayer config anyway. So let's just fix the logic.

templates/default/contentlayer.config.js
olayway: and the config is already available in contentlayer config (imported as `siteConfig`)

olayway:
@PhilippeduPreez can we close this PR? 🙂

PR #159, opened 2022-09-12
remark-smartypants added

PhilippeduPreez:
-- becomes ndash
--- becomes mdash
... becomes ellipse (even if there are spaces in between)
\n --- \n still creates a hr

### DIFF
--- a/templates/default/contentlayer.config.js
+++ b/templates/default/contentlayer.config.js
@@ -10,6 +10,7 @@ import remarkMath from 'remark-math';
 import remarkToc from 'remark-toc';
 import wikiLinkPlugin from 'remark-wiki-link-plus';
 import siteConfig from './config/siteConfig';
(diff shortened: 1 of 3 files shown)


templates/default/contentlayer.config.js
```diff
@@ -63,6 +64,7 @@ export default makeSource({
   mdx: {
     remarkPlugins: [
       remarkGfm,
+      [smartypants, {quotes: false, dashes: 'oldschool'}],
```
olayway, line 67: Let's also replace the quotes, they look a bit better this way :)

```suggestion
      [smartypants, {dashes: 'oldschool'}],
```

olayway (changes requested):
Looks good :) Just one minor thing.

olayway (approved):

PR #160, opened 2022-09-12
143 edit page link

### LINKED ISSUE #143: "Edit this page" link on pages
When creating a wiki-like site (or docs site) based on a public repo i want to be able to edit this page link on pages so that people (including me the editor) can quickly edit a page and make changes

NB: believe we have already implemented this on web3 so this is a backport

## Acceptance

* [x] Edit this page link at bottom of pages linking to right section on repo
* [x] Edit this page can be enabled / disabled (disabled by default) in config
* [x] Edit this page can be enabled / disabled per page via frontmatter option

## Tasks
* [x] Fork Flowershow main and create a branch for this issue
* [x] Edit this page link at bottom of pages linking to right section on repo. **added link to layouts.js**
* [x] Edit this page can be enabled / disabled (disabled by default) in config. **editLinkShow: false added**
* [x] Edit this page can be enabled / disabled per page via frontmatter option. **editLink with default: editLinkShow**
* [x] Review whether all changes work as it should

PhilippeduPreez:
- link has been moved to layout.js so any type of doc can show link
- logic changed so the default frontmatter bool value for showing link is the same value as in siteconfig

Closes #143


olayway:
@PhilippeduPreez I guess PR #158 is now redundant?

olayway:
Can you also link your PRs to relevant issues, so that they can be closed automatically after merge ? :) Here is a short info on that:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

templates/default/components/Layout.js
```diff
@@ -5,17 +5,49 @@ import siteConfig from '../config/siteConfig';
 import Header from './Nav';
 
 export default function Layout({ children, title = '' }) {
+  const { editLink, _raw } = children.props;
```
olayway, line 8: I had no idea components can access their children's props this way...weird 😅

templates/default/components/Layout.js
```diff
@@ -5,17 +5,49 @@ import siteConfig from '../config/siteConfig';
...
+          href="data:image/svg+xml,<svg xmlns=%22http://www.w3.org/2000/svg%22 viewBox=%220 0 100 100%22><text y=%22.9em%22 font-size=%2290%22>💐</text></svg>"
+        />
         <meta charSet="utf-8" />
         <meta name="viewport" content="initial-scale=1.0, width=device-width" />
       </Head>
       <div className="relative min-h-screen pb-52 dark:bg-slate-900">
         <Header />
         <main>
           {children}
+          {editUrl && siteConfig.editLinkShow && editLink && (
```
olayway, line 26: 1. We don't really have to check if `editUrl` is defined, since `_raw.sourceFilePath` will never be an empty string.
2. `siteConfig.editLinkShow` is already linked to the default value of `editLink` frontmatter field...(in contentlayer config file) So, the value of `editLink` property on this page will already be either this default value or the one set explicitly in frontmatter. It also means, that this solution won't work if you have `editLinkShow` set to `false` in config and try to change it for some page in frontmatter to `true`, since you'll end up with `editUrl && siteConfig.editLinkShow && editLink` = `true && false && true`, which evaluates to `false`.... So, we don't want to use `siteConfig.editLinkShow` here again. We just want to have `{ editLink && (`

But now I gave it another thought, there is even better solution than setting `editLink` frontmatter field default value to the one set in config file. We can do this:
```suggestion
          {(editLink ?? siteConfig.editLinkShow) && (
```

Which basically means:
If `editLink` value is defined (set in frontmatter), use this value to decide whether to display the link or not. Otherwise, use the value from the config file to decide.

templates/default/contentlayer.config.js
```diff
@@ -17,6 +17,7 @@ const sharedFields = {
   description: { type: 'string' },
   image: { type: 'string' },
   layout: { type: 'string', default: 'docs' },
+  editLink: { type: 'boolean', default: siteConfig.editLinkShow},
```
olayway, line 20: Yes, this is what I meant originally, but given the suggestion in my comment to `Layout`'s component logic, we won't need this default value here. So we would do:
```suggestion
  editLink: { type: 'boolean' },
```

olayway (changes requested):

olayway:
@PhilippeduPreez you can find out why checks are failing by clicking at `Details` on the right-hand side. You basically have to look at error logs and try to figure out what went wrong and why. Let me know if you don't find answers there. I'll be happy to help.

templates/default/config/siteConfig.js
```diff
@@ -3,6 +3,9 @@ import userConfig from '../content/config.js';
 const defaultConfig = {
   title: 'Flowershow',
   description: 'Publish your digital garden',
+  repoRoot: "https://github.com/flowershow/flowershow",
+  repoEditPath: "/edit/main/site/content/",
+  editLinkShow: false,
```
olayway, line 8: We actually do want to have it set to `true` by default :) I'd just turn it off for `index.md` page (explicitly, in frontmatter) as it won't look good on the home page imo.

olayway (changes requested):

Issue #111, opened 2022-08-23
Trial on Ecosystem v0.2


rufuspollock:
This is also a chance to trial out a sort of upgrade.

## Acceptance

* [x] ecosystem.lifeitself.us is running flowershow v0.2 and working **🚧2022-09-20 pending pull request https://github.com/life-itself/ecosystem/pull/88**
* [x] custom components are separated out
* [x] learnings
  * [x] contentlayer stuff
  * [x] components stuff
  * [x] summary of improvements we backported! 	

Key back-ported features **✅ 2022-09-14**

* [x] Some way to configure title of site in navbar e.g. `navbarTitle` attribute with `logo, text, version` options. **✅ FIXED set values with navbarTitle attribute in user's config**
* [x] If search settings not configured don't show search bar ** ✅  FIXED based on algolia config values**
* [x] Ability to only have one theme (e.g. not light and dark) => hide switcher, only have that theme **✅ FIXED**
* [x] Misc small fies

## Learnings

**Upgrading Flowershow on a site that uses contentlayer**

To address:
- components not importing in markdown - **Big setback to resolve**
- document types are modified in contentlayer
- package.json is replaced completely

## Tasks

* [x] Create script for upgrading **✅2022-08-29 created an `upgrade.sh` in `packages/flowershow`**
* [ ] Evaluate current setup of ecoystem to check for conflicts and hand merging **🚧2022-08-29 see notes and tasks below**

### Fixes to flowershow round 1

* [x] 🐛 site won't build without `nextSeo` being defined but it is not in default config setup
components/MDX.js (39:14) @ MdxPage
```
  37 |           }
  38 |         ] 
> 39 |       : siteConfig.nextSeo.openGraph.images,
     |        ^
  40 |   }}
  41 | />
  42 | <Component
```
* [x] data/exampleData.js - why do we have this? Can this live somewhere else e.g. in site? **olayway: exampleData moved to /site/data/exampleData; it makes import paths in MDX a bit ugly but I've created an issue for it - #154**
* [x] `siteConfig.js` remove commented out nextSeo stuff (since commented out so just noise)
* [x] `MDX.js` **✅ 2022-08-31**
  * [x] Backport Link
  * [x] Backport `svg: props => <Fragment {...props} />`
* [x] `Nav.js` We have hard-coded github icon but that should be optional **✅ 2022-08-31**
[...]

PR #161, opened 2022-09-12
[#111]: Backport changes from ecosystem

khalilcodes:
## Backport changes to flowershow after ecosystem upgrade
Tasks from #111 
***

This PR adds changes in  User config (for version), Navbar title, Dark/Light theme switcher, Search component and flowershow image assets.

### Tasks

- [x] Create `_flowershow` folder in public with images (logo and theme icon)
- [x] ~~Add `navbarTitle` object in `siteConfig.js` specific to flowershow stuff (title text, version, logo, etc)~~ **revert changes**
- [x] Create Navbar title function in `components/Nav,js` to render site title based off config values
  - [x] values are from user's `config.js` and properties can be set in optional `navbarTitle` for `text [String]`, `logo [Boolean]` and `version [Boolean]`
  - [x] ~~handle SEO title for both cases in `pages/_app.js`~~ **revert to previous config**
- [x] Add conditional rendering for Search component if docsearch configs exist (ie., appId, apiKey and indexName)
  - [x] Add a `env.development` file with test values to render search component in development mode
- [x] Add forced light theme and handle rendering of theme button based on config ie., if `theme.default` is an empty string
- [x] Amend name of url path field to `url_path` in. `contentlayer.config.js` to avoid conflicts if any

### Notes
- ~~Searchbar component may not render in preview deploy as env variables will not be loaded. The searchbar rendering relies on having all three required algolia configs to be defined (ie. appId, apiKey and indexName)~~ **This works**
- `upgrade.sh` package needs to be updated to remove `.env.development` file


site/content/config.js
```diff
@@ -1,5 +1,5 @@
 const config = {
-  title: 'Flowershow',
+  navbarTitle: { version: true },
```
rufuspollock, line 2: surely `navbarTitle` is something like:

```
{
	title: 'Flowershow',
	logo: ...
	version: 'Alpha'
}
```

templates/default/components/Nav.js
```diff
@@ -14,6 +14,36 @@ function GitHubIcon(props) {
   );
 }
 
+function NavbarTitle() {
+  // include option to define different text from title (if needed) in config
+  const title = siteConfig.navbarTitle?.text || siteConfig.title
```
rufuspollock, line 19: Personally, i would be strict and say if you use navbarTitle you have to set title in their too i.e. this is:

```
if (navbarTitle) {
	use stuff from navbarTitle
} else {
      use siteConfig.title
}
```

Why? Explicitness is better than magic!

rufuspollock (approved):
I think code is ok for now though i have a few comments (don't need to change for them). Big thing missing is a bit of docs. I can do this quickly.

PR #180, opened 2022-09-23
178-playwright-features

### LINKED ISSUE #178: Feature Testing using playwright round 2 part II: more tests with new fixtures
We want to test more features of the template app so that as we add things we know we aren't breaking anything

## Acceptance

* [x] Key features of template are under test

## Tasks

Select more stuff to test (and e.g. philippe can have a go)

* [x] Review main features and pick a few key things we want to test
* [x] Syntax
  * [x] #118 
  * [x] wikilinks
  * [x] CommonMark
  * [x] GFM

PhilippeduPreez:
fixes #178 
added playwright testing for features including commonmark, GFM, wikilinks etc


rufuspollock:
@PhilippeduPreez @olayway just to say i did a quick review and good to go IMO.

rufuspollock:
@PhilippeduPreez can you fix the conflicts so we can merge.

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
+const { test, expect } = require('@playwright/test');
+const { MarkdownPage } = require('./markdown-page');
+
+test.describe('wiki links', () => {
```
olayway, line 4:
```suggestion
test.describe.parallel('wiki links', () => {
```

You can add `.parallel` to every test group so that the tests grouped within it don't run one after the other but in parallel. It will remove the warning that's currently shown in the console about the slow test files - and will basically speed up the process :)

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  test.beforeEach(async ({ page }) => {
+    const Page = new MarkdownPage(page);
+    await Page.goto('/test/fixtures/markdownFeatures');
+    await Page.getData(page);
```
olayway, line 8: You shouldn't have to pass `page` again to this method. It's already saved on the `MarkdownPage` instance on creation. Let's fix this method so that it uses `this.page` instead.

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  test('parses a wiki link', async ({ page }) => {
+    const link = page.locator('#wikiLink > p > a');
+    await expect(link).toHaveClass('internal');
+    await link.click();
+    await expect(page).toHaveURL('/docs/roadmap');
```
olayway, line 15: This test will fail if we move the roadmap page to some other location or we just delete it. Tests should be totally decoupled from the actual content so that they don't rely on things that may easily change in the future. => Let's create a fixture page for this purpose and link to it in `markdownFeatures.md`.

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  test('parses a wiki link with custom divider', async ({ page }) => {
+    const link = page.locator('#wikiLink-custom > p > a');
+    await expect(link).toHaveClass('internal');
+    await link.click();
+    await expect(page).toHaveURL('/docs/roadmap');
```
olayway, line 22: as above

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  test('parses a wiki link with header', async ({ page }) => {
+    const link = page.locator('#wikiLink-heading > p > a');
+    await expect(link).toHaveClass('internal');
+    await link.click();
+    await expect(page).toHaveURL('/docs/roadmap#planned-features');
```
olayway, line 29: as above

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  test('parses a wiki link with header and custom divider', async ({ page }) => {
+    const link = page.locator('#wikiLink-heading-custom > p > a');
+    await expect(link).toHaveClass('internal');
+    await link.click();
+    await expect(page).toHaveURL('/docs/roadmap#planned-features');
```
olayway, line 36: as above

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  test('link to image file', async ({ page }) => {
+    const link = page.locator('#wikiLink-image > p > img');
+    await expect(link).toHaveClass('internal');
+    await expect(link).toHaveAttribute('src', '/assets/images/park.png');
```
olayway, line 42: This will also break if we remove the `park.png` image when we don't need it anymore. Let's copy this image to where the markdown fixtures are (maybe in some `/assets` subfolder for now, as I'm not sure what is the best practice here)

olayway (changes requested):
Great job! Some minor changes to enhance performance and ensure the tests won't break in the future.

olayway:
@PhilippeduPreez, @rufuspollock I've also checked that it's not necessary to manually run the app (`npm run dev`) before we can run tests. Playwright will do this automatically in the background. We just didn't have the correct config file.

This is the config I've tested: https://github.com/flowershow/flowershow/blob/test-playwright/templates/default/playwright.config.mjs (it's a slightly adjusted version of the file taken from Next.js' examples)

I've also found a GH CI configuration in official Playwright docs and have also adjusted it a bit - https://github.com/flowershow/flowershow/blob/test-playwright/.github/workflows/test-e2e-app.yml.

Tested both on [this branch](https://github.com/flowershow/flowershow/tree/test-playwright). I'll create a PR once this PR is merged.

templates/default/tests/markdownFeatures.spec.js
PhilippeduPreez: I added a testimages subfolder to images for now, but we can change the location if there is better practice for this.

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
...
+    await link.click();
+    await expect(page).toHaveURL('/test/fixtures/fixturepage');
+  });
+
+  test('parses a wiki link with custom divider', async ({ page }) => {
+    const link = page.locator('#wikiLink-custom > p > a');
+    await expect(link).toHaveClass('internal');
+    await link.click();
+    await expect(page).toHaveURL('/test/fixtures/fixturepage');
+  });
```
olayway, line 23: These two tests do exactly the same thing. The second one should also check if the text inside of the link is the custom one (written after `|`).

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
...
+    await link.click();
+    await expect(page).toHaveURL('/test/fixtures/fixturepage#planned-features');
+  });
+
+  test('parses a wiki link with header and custom divider', async ({ page }) => {
+    const link = page.locator('#wikiLink-heading-custom > p > a');
+    await expect(link).toHaveClass('internal');
+    await link.click();
+    await expect(page).toHaveURL('/test/fixtures/fixturepage#planned-features');
+  });
```
olayway, line 37: same here

templates/default/tests/markdownFeatures.spec.js
```diff
@@ -0,0 +1,136 @@
...
+    await expect(page.locator('#lists > ul > li')).toContainText('one');
+    await expect(page.locator('#lists > ol > li')).toContainText('one');
+    await expect(page.locator('#lists > ol > li > ul > li')).toContainText('one');
+  });
+
+  test('links', async ({ page }) => {
+    const link = page.locator('#links > p > a');
+    await expect(link).toContainText('Link to roadmap');
+    await link.click();
+    await expect(page).toHaveURL('/docs/roadmap');
```
olayway, line 105: this may also break in the future

olayway (changes requested):

templates/default/.husky/post-checkout
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting '.git/hooks/post-checkout'.\n"; exit 2; }
+git lfs post-checkout "$@"
```
olayway, line 3: What is this file doing here?

templates/default/.husky/post-commit
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting '.git/hooks/post-commit'.\n"; exit 2; }
+git lfs post-commit "$@"
```
olayway, line 3: What is this file doing here?

templates/default/.husky/post-merge
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting '.git/hooks/post-merge'.\n"; exit 2; }
+git lfs post-merge "$@"
```
olayway, line 3: What is this file doing here?

templates/default/.husky/pre-push
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting '.git/hooks/pre-push'.\n"; exit 2; }
+git lfs pre-push "$@"
```
olayway, line 3: What is this file doing here?

PhilippeduPreez: I was actually also confused where that came from?

.changeset/olive-apes-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@fake-scope/fake-pkg": patch
+---
+
+playwright testing addded for features to make sure nothing breaks as the project grows
```
olayway, line 5: This file shouldn't be placed here, it should go into `templates/default/.changeset`

olayway (changes requested):

olayway (approved):
The tests work as expected locally so the issue we have here is probably related to our current playwright setup. I'll address this in another PR.

PR #182, opened 2022-09-23
Fix components imports in MDX files

### LINKED ISSUE #157: Importing components (in MDX) that use external packages doesn't work
@khalilcodes, @rufuspollock it seems importing components that depend on external packages doesn't work indeed 😞 I believe it has to do with Contentlayer's configuration/usage of `mdx-bundler`. When you actually define the component and import it's dependencies directly in MDX file (not import them from a file) everything works. So it has sth to do with imports resolution I guess. Anyways, I've created an issue in Contentlayer's repo for that: https://github.com/contentlayerdev/contentlayer/issues/302

You can also test it yourself. I've added relevant tests  to `/content/test/components-import.md` so that Contentlayer's team can reproduce the issue.

I think we should also get acquainted with Contentlayer's source code sometime, and also with the docs of `mdx-bundler`', which they are using under the hood, so that it's not a complete black box for us.

_Originally posted by @olayway in https://github.com/flowershow/flowershow/issues/111#issuecomment-1241198848_

olayway:
Closes #157

## Changes
- upgrade contentlayer version to 0.2.8
- change Contentlayer's default cwd (in 0.2.8 it's the parent directory of each document) to `process.cwd()`, which resolves to `templates/default` directory in our case
- update import paths 

## Notes
Unfortunately, it doesn't solve the problem with Next's Link. This is a totally different type of error.


rufuspollock (approved):

rufuspollock:
@olayway it looks like we need to resolve conflicts - probably due to eslint work.

PR #196, opened 2022-10-06
128 private notes

PhilippeduPreez:
add any files/dir in contentExclude config that should not be published
isDraft: true can be added in frontmatter in order not to be published, even if it is in a dir that is published

### DIFF
--- a/site/content/config.js
+++ b/site/content/config.js
@@ -13,6 +13,7 @@ export const config = {
   repoEditPath: "/edit/main/site/content/",
   editLinkShow: false,
   contentExclude: [],
+  contentInclude: [],
   analytics: "G-RQWLTRWBS2",
   navLinks: [
     { href: "/#overview", name: "Overview" },
--- a/site/content/config.js
+++ b/site/content/config.js
@@ -12,6 +12,7 @@ export const config = {
   repoRoot: "https://github.com/flowershow/flowershow",
   repoEditPath: "/edit/main/site/content/",
   editLinkShow: false,
+  contentExclude: [],
   analytics: "G-RQWLTRWBS2",
   navLinks: [
     { href: "/#overview", name: "Overview" },
--- a/site/content/docs/syntax.md
+++ b/site/content/docs/syntax.md
@@ -301,6 +301,28 @@ Two '-' will convert to ndash. Three '-' will convert to mdash. Three '.' with o
 ...ellipse\
 ...another ellipse
 
+### ✅ Callouts
+
+Flowershow supports 12 different Obsidian callout types (with aliases) like note, abstract, todo, or tip. See this [Obsidian docs page](https://help.obsidian.md/How+to/Use+callouts) to learn more about different callout types.
+
+**Example:**
+
+```md
+> [!info]
+> Here's a callout block.
+> It supports **markdown** and [[Internal link|wikilinks]].
+```
+
+**Renders as:**
+
+> [!info] This is cool!
+> Here's a callout block.
+> It supports **markdown** and [[docs/index|wikilinks]].
+
+<div className="border-2 border-slate-400 rounded-md px-4 my-4">
+🔍 To learn more about the Obsidian extensions refer to the [Obsidian Help site](https://help.obsidian.md/How+to/Format+your+notes). 
+</div>
+
 ### 🚧 PDF embedding
 
 **Example:**
@@ -315,24 +337,6 @@ Two '-' will convert to ndash. Three '-' will convert to mdash. Three '.' with o
 ### 🚧 Text highlighting
 
 ==I'm Highlighted!== is done using `==I'm Highlighted!==`
-
-### 🚧 Callouts
-
-**Example:**
-
-```md
-> [!INFO]
-> Here's a callout block.
-> It supports **markdown** and [[Internal link|wikilinks]].
-```
-
-**Renders as:**
-
-> [!INFO] > Here's a callout block. > It supports **markdown** and [[docs/index|wikilinks]].
-
-<div className="border-2 border-slate-400 rounded-md px-4 my-4">
-🔍 To learn more about the Obsidian extensions refer to the [Obsidian Help site](https://help.obsidian.md/How+to/Format+your+notes). 
-</div>
 
 ### 🚧 Excalidraw sketches support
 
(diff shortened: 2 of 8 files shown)


templates/default/.changeset/seven-starfishes-march.md
```diff
@@ -0,0 +1,5 @@
+---
+"default": patch
+---
+
+option in config to add files/dir not to publish
```
olayway, line 5: we only need one changeset file

site/content/config.js
```diff
@@ -12,6 +12,7 @@ export const config = {
   repoRoot: "https://github.com/flowershow/flowershow",
   repoEditPath: "/edit/main/site/content/",
   editLinkShow: false,
+  contentExclude: [],
```
olayway, line 15: It would be good to also utilize the other contentlayer option: `contentDirInclude`, so that you can have all content excluded by default except for the small set of specific folders/files.

templates/default/config/siteConfig.js
```diff
@@ -16,6 +16,7 @@ const defaultConfig = {
   // if you have your notes in another (external) directory,
   // /content dir should be a symlink to that directory
   content: "content",
+  contentExclude: ["fake"],
```
olayway, line 19: I guess this is some test value we don't need?

site/content/config.js
olayway: You will also have to test what happens if you set both these values. I can't see any info on that in contentlayer's docs so I think it may work just fine, i.e. if you select a given subdirectory you want to publish and you want to exclude some child directory/file, e.g. include `content/docs` and exclude `content/docs/config.mdx`

olayway (changes requested):
Also, don't forget to add a section about it to the `docs/config` page.

templates/default/config/siteConfig.js
PhilippeduPreez: Indeed :)

site/content/config.js
PhilippeduPreez: It works when including a folder and excluding a file in that folder

PhilippeduPreez:
image
@olayway I am unsure why I am getting this error?

olayway:
@PhilippeduPreez I'm not sure. I've tried to reproduce it, and I only managed to do it once. Now it works.. But still there are some things you're doing in your code which are not necessary.

templates/default/config/siteConfig.js
```diff
@@ -29,6 +31,8 @@ const defaultConfig = {
 export const siteConfig = {
   ...defaultConfig,
   ...config,
+  contentExclude: [...defaultConfig.contentExclude, ...config.contentExclude],
```
olayway, line 34: It's not necessary. The behavior we want to achieve is the ability to set/override contentExclude or contentInclude by the user.

templates/default/contentlayer.config.js
```diff
@@ -77,7 +78,9 @@ export default makeSource({
   contentDirExclude: contentLayerExcludeDefaults.concat([
     ".flowershow",
     ".obsidian",
+    ...siteConfig.contentExclude,
   ]),
+  contentDirInclude: [...siteConfig.contentInclude],
```
olayway, line 83: if `siteConfig.contentInclude` is an array, doing `[...siteConfig.contentInclude]` will create a copy of it, which is not necessary.

templates/default/pages/[[...slug]].jsx
```diff
@@ -20,6 +21,9 @@ export async function getStaticProps({ params }) {
 export async function getStaticPaths() {
   const paths = allDocuments.map((page) => {
     const parts = page.url_path.split("/");
+    if (page.isDraft) {
+      return { params: { slug: undefined } };
+    }
     return { params: { slug: parts } };
   });
```
Comment (olayway) [inline templates/default/pages/[[...slug]].jsx:28]: Wouldn't it be better to simply filtered out pages for which `page.isDraft` is set to `true`?

site/content/docs/syntax.md
```diff
@@ -301,38 +301,42 @@ Two '-' will convert to ndash. Three '-' will convert to mdash. Three '.' with o
...
 
-> [!INFO] > Here's a callout block. > It supports **markdown** and [[docs/index|wikilinks]].
+### 🚧 Text highlighting
 
-<div className="border-2 border-slate-400 rounded-md px-4 my-4">
-🔍 To learn more about the Obsidian extensions refer to the [Obsidian Help site](https://help.obsidian.md/How+to/Format+your+notes). 
-</div>
+==I'm Highlighted!== is done using `==I'm Highlighted!==`
 
 ### 🚧 Excalidraw sketches support
```
olayway, line 342: what's going on here? :D I suspect this wasn't intended for this PR ;)

site/content/index.md
```diff
@@ -176,6 +176,12 @@ Support for for meta tags for SEO and social sharing.
...
 
+### ✅ CLI tool
+
+Our goal is to make using Flowershow as seamless as possible. To facilitate smooth bootsrapping and upgrading your website we're creating a CLI tool, that will take care of all the intricacies related to the whole process of publishing your notes.
+
+[[publish-tutorial|Learn more ➡]]
+
 ### 🚧 Knowledge graph
 
 Interactive knowledge graph to see how your notes are interconnected.
```
olayway, line 187: what's going on here? :D I suspect this wasn't intended for this PR ;)

site/content/index.md
```diff
@@ -196,10 +202,6 @@ Lists of forward links used on the page and back links to the page from other pa
 
 Displaying Excalidraw sketches directly in your markdown pages.
 
-### 🚧 CLI tool
-
-Our goal is to make using Flowershow as seamless as possible. To facilitate smooth bootsrapping and upgrading your website we're creating a CLI tool, that will take care of all the intricacies related to the whole process of publishing your notes.
-
     👷 There is a lot of other exciting stuff we're working on.
     <a href="/docs/roadmap">Check our roadmap to learn more!</a>
```
olayway, line 207: what's going on here? :D I suspect this wasn't intended for this PR ;)

olayway (changes requested):
Philippe, please, each time you do a commit, check what you have in your staging area ;) I used to do the same kind of mistakes, because of doing things too fast and not double checking my commits, but it is a really good habit. It will save you a lot of headaches, trust me ;)

PR #210, opened 2022-10-18
RHS Table of Contents

### LINKED ISSUE #95: RHS ToC (sidebar current page table of contents)
A user can quickly eyeball the contents of the current page by looking at the table of contents (ToC) on the right-hand side. He can also quickly navigate to different sections of the current page by clicking on a section title in the ToC and see where he's currently at thanks to highlighting of the currently focused section.

## Acceptance

- [x] RHS table of contents shows content from current page (excluding the main/page heading??)
- [x] Updates as we scroll the page
- [x] Highlights as we reach different sections
- [x] Questions
  - [x] What happens on non-markdown pages? **Doesn't show**
  - [x] Can we disable toc display? **Yes, in `config.js` or frontmatter see [docs/table-of-contents](https://flowershow.app/docs/table-of-contents) for more info**

## Tasks
- [x] Add ToC Component to layout
- [x] Add `toc` frontmatter field (boolean)
  - [x] set to false on homepage
- [x] Add `tableOfContents` boolean in `config.js`
  - ~~supports all pages (except homepage/non-markdown pages) `tableOfContents: true`~~
  - ~~supports for only specific page/folders on content dir eg. `tableOfContents: ["docs", "about"]`~~
- [x] Add instructions for RHS ToC to docs/table-of-contents

## Screenshot
Screen Shot 2022-10-18 at 2 47 02 AM

## Notes
Tailwind's solution analysis: https://github.com/datopian/product/blob/main/notes/tailwindcss-website-toc.md

khalilcodes:
## Support Table of contents on RHS for markdown pages
Closes #95 
***

This PR adds the ToC sidebar (RHS) feature to markdown pages by setting in config.

### Tasks
- [x] Create functions and hooks for ToC headings tree
- [x] Add ToC Component to layout
- [x] Add `tableOfContents` in config.js **UPDATED boolean value only**
  - ~~add support for all pages (except homepage/non-markdown pages) `tableOfContents: true`~~
  -  ~~add support for only specific page/folders on content dir eg. `tableOfContents: ["docs", "about"]`~~
- [x] Add `toc` frontmatter field (boolean)
  - [x] set this to false on homepage
- [x] Update docs add instructions for RHS ToC - [docs/table-of-contents](https://deploy-preview-210--spectacular-dragon-c1015c.netlify.app/docs/table-of-contents)

### Screenshot

Screen Shot 2022-10-18 at 2 47 02 AM


rufuspollock:
@olayway are you happy to review this?

templates/default/pages/_app.jsx
```diff
@@ -13,6 +13,34 @@ import "../styles/docsearch.css";
...
+        }
+      }
+    }
+
+    return sections.push(...collectHeadings(node.children ?? []));
+  });
+
+  return sections;
+}
+
```
olayway, line 43: What if we:
1. Replace `map` with `forEach`. This is more suitable for this use case as we don't need to create a new array, we just want to iterate over an array and do sth for some nodes. This way you also don't need to create artificial return statements in each if block.
2. Try to simplify the code a bit so that we don't have so many nested if blocks as it is really hard to reason about.

```suggestion
  Array.from(nodes).forEach((node) => {
    const { id, innerText: title, tagName: level } = node;
    if (!(id && title)) {
      return;
    }
    if (level === "H3") {
      const parentSection = sections[sections.length - 1];
      parentSection && parentSection.children.push({ id, title });
    } else if (level === "H2") {
      sections.push({ id, title, children: [] });
    }
    
    sections.push(...collectHeadings(node.children ?? []));
  });

  return sections;
}

```

templates/default/pages/_app.jsx
```diff
@@ -28,6 +56,14 @@ function MyApp({ Component, pageProps }) {
     }
   }, [router.events]);
 
+  const [tableOfContents, setTableOfContents] = useState([]);
+
+  useEffect(() => {
+    const headingNodes = document.querySelectorAll("h1,h2,h3,h4,h5,h6");
```
olayway, line 62: I guess we could only look for `h2` and `h3` as we don't use the rest of them?

site/content/docs/table-of-contents.md
```diff
@@ -31,3 +31,34 @@ The table of contents will only include headings placed below the `Table of cont
...
+### ToC for custom pages/folders
+
+Sometimes there is a need for table of contents only on specific pages, for example docs or blog pages. Flowershow also supports enabling ToC only for these specific defined page/pages by passing that directory or path from your content folder in an array as seen below.
+
+```js
+// config.js
+
+const config = {
+  tableOfContents: ["docs", "blog", "about"]
+}
```
olayway, line 61: Let's keep this always of the same type. It's rather strange that it can be either boolean or list of strings. Also, up until know we used config to set default values that could later be changed in frontmatter for specific pages. Let's maybe stick with this approach?

templates/default/components/Layout.jsx
```diff
@@ -1,13 +1,76 @@
...
+
+export function Layout({ children, tableOfContents }) {
   const { editLink, _raw } = children.props;
   /* if editLink is not set in page frontmatter, link bool value will depend on siteConfig.editLinkShow */
   const editUrl = `${siteConfig.repoRoot}${siteConfig.repoEditPath}${_raw?.sourceFilePath}`;
+
+  const showToc = Array.isArray(siteConfig.tableOfContents)
+    ? siteConfig.tableOfContents.includes(_raw?.sourceFileDir) ||
+      siteConfig.tableOfContents.includes(_raw?.flattenedPath)
+    : _raw?.flattenedPath;
```
olayway, line 60: So `showToc` can either be a boolean or a string ? (_raw?.flattenedPath). I think I don't understand :)

templates/default/components/Layout.jsx
```diff
@@ -1,13 +1,76 @@
...
+        const el = document.getElementById(id);
+        if (!el) return;
+
+        const style = window.getComputedStyle(el);
+        const scrollMt = parseFloat(style.scrollMarginTop);
+
+        const top = window.scrollY + el.getBoundingClientRect().top - scrollMt;
+        return { id, top };
+      });
+  }, []);
```
olayway, line 24: Does `[]` here means it will never be cached or it will always be cached? Does empty array do anything?

olayway (changes requested):
Works great but I've added a few suggestions and questions.

templates/default/components/Layout.jsx
khalilcodes: _raw?.flattenedPath for index.md (ie. frontpage) returns an empty string and because of this the ToC wont render on the homepage. It's not a boolean but being used like one.

site/content/docs/table-of-contents.md
khalilcodes: So does the following look correct maybe ?

```js
const config = {
  tableOfContents: ["all"]
}

File: templates/default/components/Layout.jsx (hunk 5 above)
Comment (khalilcodes) [reply templates/default/components/Layout.jsx:24]: Basically we are telling here to compute these values once since we just need them initially or else they will keep regenerating with the scroll top which makes the function expensive.

File: site/content/docs/table-of-contents.md (hunk 3 above)
Comment (khalilcodes) [reply site/content/docs/table-of-contents.md:61]: @olayway I may not have understood here. Do we want to be able to have something like below in frontmatter for specific pages ?
```md
---
tableOfContents: true
---
```

olayway: I'd do it the same way we handle e.g. showing edit this page link, wich is enabling/disabling by default using boolean value in config, with the possibility to override this value in frontmatter.

templates/default/components/Layout.jsx
olayway: Yes, I figured that you're using this string value as a truthy/falsy value but it's hard to understand what you're trying to achieve for someone who doesn't have the explanation. Also, if the `siteConfig.tableOfContents` is not an array, you're completely ignoring it's boolean value which can be either `true` or `false`, because all the pages apart from the index page will not be an empty string. So `showToc` can evaluate to the truth value even though you have `tableOfContents` set to `false` in the config.

khalilcodes: Right will make those changes 👍

templates/default/pages/_app.jsx
khalilcodes: Have tested and this works. One minor change I had to add an if statement on line 44 as esbuild was throwing error during build `Expected an assignment or function call and instead saw an expression.`

templates/default/components/Layout.jsx
```diff
@@ -29,14 +29,12 @@ function useTableOfContents(tableOfContents) {
     function onScroll() {
       const top = window.scrollY + 4.5;
       let current = headings[0].id;
-      // for (const heading of headings) {
       headings.map((heading) => {
         if (top >= heading.top) {
           current = heading.id;
         }
         return current;
       });
```
olayway, line 37: Again, we don't need a `map` here as we only need to iterate over this array and not create a new one.

site/content/docs/table-of-contents.md
```diff
@@ -31,3 +31,32 @@ The table of contents will only include headings placed below the `Table of cont
...
+***
+
+## Right Hand Side (RHS)
+
+Flowershow supports adding a Table of Contents (ToC) sidebar (RHS) to your pages which can be easily setup in your config file.
+
+![[rhs-toc.jpg]]
+
+### ToC on all pages
+To enable ToC on all the markdown pages (except homepage), simply set the table of contents variable to true in your config file as seen below:
```
olayway, line 44: Let's remove "except homepage" now we actually support it as well

olayway (approved):
LGTM :) just some minor things

olayway:
@khalilcodes can you merge this?

PR #221, opened 2022-10-24
Backport import components

olayway:
## Summary

Support for using data from `contentlayer/generated` in markdown pages (that themselves are included in `contentlayer/generated`). This data is obtained in `getStaticProps`, which is called before next build but after contentlayer build, which means all the data from `contentlayer/generated` can be accessed and passed to consumer pages, and further to components used in them.
Consumer pages can define variables they need access to (e.g. to pass them to components) in their frontmatter, in newly added `data` field. This field is of type `list<string>`, which means you can define multiple variables you want to have access to.
Names of the variables in the list need to have a corresponding getter function created in `content/getters` folder.
These getters will be called by `getPageData` function in `getStaticProps`, so after `contentlayer/generated` has been populated. This means all getter functions can import from `contentlayer/generated`.
A page with names defined in frontmatter's `data` field will have access to variables under the same names (provided that corresponding getters exist).

## Changes
- create `getPageData` function that takes in an array of variable names <string>, for each name it tries to import a getter of a corresponding name from `content/getters`, obtain the data, and return an object with a mapping from each name -> data returned from the getter
- define a new contentlayer field: `data` of type list<string> (default: [])
- get value of `data` field  for each MD page in `getStaticProps` and call `getPageData` with it
- pass the resulting object with mapping from each name from `data` field list to the result of running a corresponding getter function, as `data` prop to `Page` component in `[[...slug]]`
- pass `data` object to MDX page

## Example
Some MD page, which requires profiles and topics variables to render its `CircularVis` component.

```md
---
data:
  - profiles
  - topics
---

import CircularVis from "components/ecosystem/CircularVis.jsx"
## Visualizing the Ecosystem by Topic

This visualization maps an overview of organizations grouped together by their primary topic. Click to zoom in or out. Hover over a circle to see its name.

<CircularVis profiles={profiles} topics={topics} />
```

We need to create corresponding `content/getters/profiles.js` and `content/getters/topics.js`, where we can import from `contentlayer/generated` E.g.:
```js
// profiles.js
import { allProfiles } from 'contentlayer/generated';

export default function getProfiles() {
  return allProfiles.filter(
    (profile) => !(profile.curation_status.includes('N') || profile.curation_status.includes('?'))
  );
}

```


rufuspollock:
@olayway PR description is excellent (can become the docs i suspect!).

One minor comment about the PR is that we have a bit of noise from the linting stuff - in a perfect world that would be a separate PR (or just push those fixes to main directly) so that this PR is clean. Ok in the end as i looked at commits and commits are clean.

templates/default/pages/[[...slug]].jsx
```diff
@@ -3,9 +3,10 @@ import { allDocuments } from "contentlayer/generated";
 import { useMDXComponent } from "next-contentlayer/hooks";
 
 import { MdxPage } from "../components/MDX";
+import { getPageData } from "../lib/getPageData";
 
-export default function Page({ body, ...rest }) {
-  const mdxComponent = useMDXComponent(body.code);
+export default function Page({ body, data, ...rest }) {
```
Comment (rufuspollock) [inline templates/default/pages/[[...slug]].jsx:8]: What was `...rest` before this?

One general question: does a given page have access to its frontmatter as data in the page.

rufuspollock (approved):
Very simple afaict so looks good. I think some documentation will be important though that can be a separate PR once we have tried this out with ecosystem and got any bugs ironed out.

templates/default/lib/getPageData.js
```diff
@@ -0,0 +1,15 @@
+export async function getPageData(dataProperties) {
+  const data = {};
+  for (const dataProperty of dataProperties) {
+    try {
+      const dataGetter = (await import(`../content/getters/${dataProperty}.js`))
```
rufuspollock, line 5: this is cool - i didn't realize one could do dynamic imports like this. This may help us refactor `siteConfig.js` to have a dynamic import location for the `config.js` - see https://github.com/flowershow/flowershow/issues/7

rufuspollock:
@olayway good to go IMO - just a couple of comments for discussion and my understanding of the code.

templates/default/pages/[[...slug]].jsx
Comment (olayway) [reply templates/default/pages/[[...slug]].jsx:8]: `...rest` hasn't changed after adding `data` property. It's still just "the rest" of properties. In this case it holds all the properties off the corresponding JSON file created by Contentlayer for this page, apart from `body`.

> One general question: does a given page have access to its frontmatter as data in the page.
No, afaik it doesn't.

PR #241, opened 2022-11-02
Monorepo setup

### LINKED ISSUE #18: Monorepo setup and scaffolding
We will have various things in here probably so let's do the monorepo setup.

## Acceptance

* [x] Monorepo setup with `flowershow` package in it at `packages/flowershow` **🚧2022-10-25 ❓ think we could rename CLI to be named flowershow consistent with npm naming**  **✅ 2022-10-31 olayway: from what I see in how nx handles creating new packages/apps by their generators in monorepos, how they do this in their docs but also how community does it, it's best to have all our packages published under flowershow organization, and publish them as scoped packages, i.e. they will be available under prefix `@flowershow`. Cool thing about it is that we can name our packages whatever we want and we don't have to worry about name clashes with some other public packages, so we can have our CLI package named just `cli`, and it will be published under `@flowershow/cli`**
* [x] Moved `templates/default` into `packages` e.g. at `packages/template` (we can drop default for now!)
* [x] Linting and prettier in right place
* [x] tests in right place
* [x] changeset setup documented for monorepos (in e.g. flowershow.app/docs/developers

### Preliminaries

* [x] Compare briefly monorepo tools e.g. nx, lerna, turborepo and make a choice
  * [ ] Brief summary of each
  * [x] Choice **✅2022-07-01 Let's go with nx**
* [x] Create `packages/` directory **✅2022-08-?? done**
* [x] Setup nx **🚧2022-10-25 ❓ is this relevant anymore?**
* [x] Briefly document for other devs about anything they need to do in e.g. base README.md 

### Doing the work

* [x] Monorepo setup with `flowershow` package in it at `packages/flowershow` **🚧2022-10-25 ❓ think we could rename CLI to be named flowershow consistent with npm naming**
* [x] Moved `templates/default` into `packages` e.g. at `packages/template` (we can drop default for now!)
* [x] Linting and prettier in right place
* [x] changeset setup documented for monorepos (in e.g. flowershow.app/docs/developers

olayway:
Closes #18

## Changes
- adds Nx to the repo,
- moves `templates/default` to `packages/template`
- creates `packages/template-e2e` package for template e2e tests
  - also: moves playwright dependency to the root, so that it can be used for other projects in the future
  - also: changes the bash setup script to use temporary directory, where template and content fixture are copied and symlinked, so that setup script doesn't ever touch the original template codebase and we don't need to clean up (i.e. revert symlinks to original ones) after the tests have run,
- configures Nx Cloud distributed caching, so that devs can use cached results of tasks run by others
- moves eslint dependency and base config to the root
  - each project can also have its own eslint.json file for project-specific configurations
- moves lint-staged, husky and changeset to the root
- adds basic info related to Nx and changesets in `README.dev.md`
- adds a note about Nx basics in `content/notes/Nx.md`
- renames the CLI to namespaced `@flowershow/cli` 

**ALSO**

- adds docs section about recommended package manager - pnpm - and basic configurations
  - npm workspaces which were added to support multi-package changesets setup in the monorepo have broken template's dependency resolution due to npm's hoisting algorithm
  - there is no option to opt-out of hoisting in npm workspaces
    - pnpm has this option but it turns out their algorithm correctly installs all the dependencies, so turning hoisting off is not even needed
    - BONUS: it's much faster than npm

## Suggestions

1. I think we should also include all the other packages that we currently have outside of this repository but are using them as dependencies, i.e. all three remark plugins.
  - It would make it much easier from the developer's perspective. Since everything is in the same repository there is a guarantee that all the changes you make will work together, if they work together locally.
  - You don't need to make some workaround solutions for local development of such dependencies, to test if the changes you make work with the consumer package (e.g. change `package.json` to point to the local directory with the dependency project you're working on)
  - You don't need to remember about updating versions of the dependencies, as consumer packages will always use the current code found in the monorepo

2. Create an app in `/apps` with current flowershow website (i.e. a project running off of the flowershow template with /site folder inside)
- currently the `/template` package is a hybrid of package/app, 
- we could then have a separate changesets and changelogs for our site's custom components, configs, site content etc.
- (this is probably for the future, when we have the upgrade process figured out)

3. Publish all packages under @flowershow organization

## TODOs
- [ ] publish CLI to npm
- [ ] update deploy configs on Netlify with the new location of the template


olayway:
@rufuspollock the E2E tests and builds won't succeed until this is merged, the new version of CLI is published, and deploy settings on Netlify are adjusted. I'll do all of it once this is approved.

rufuspollock:
## Minor

> adds basic info related to Nx and changesets in README.dev.md

I'd say we move this to flowershow docs e.g. flowershow.app/docs/contributors or just have a section in main README.md on this (i'd prefer the former a little).

> renames the CLI to namespaced @flowershow/cli

to check i can still do `npx flowershow` right? I much prefer that to `npx flowershow@cli xyz`?

> I think we should also include all the other packages that we currently have outside of this repository but are using them as dependencies, i.e. all three remark plugins.

Generally agree on this though as discussed yesterday YAGNI means let's see how things go for next few weeks and then decide if worth merging.

> Create an app in /apps with current flowershow website (i.e. a project running off of the flowershow template with /site folder inside)

This is a good idea and something i think we discussed. Suggest we open an issue for this where you can set out in a bit more detail and we can discuss

README.dev.md
```diff
@@ -0,0 +1,186 @@
+# Developer docs
+
+## Package manager
+
+To ensure correct dependencies installation across the projects (and for other benefits like much faster installs), please use `pnpm` package manager.
```
rufuspollock, line 5: Why do we need to pnpm - and is this a requirement or a suggestion?

README.dev.md
```diff
@@ -0,0 +1,186 @@
+# Developer docs
```
rufuspollock, line 1: I think a short quick start here explaining what is needed day-to-day for key tasks would be useful. I imagine key point is:

- Adding a feature (doing a branch + PR)
- Doing a "release"

site/content/notes/Nx.md
```diff
@@ -0,0 +1,225 @@
+# Nx
```
rufuspollock, line 1: Normally all files have lower-case names. I think nextjs automatically lower-cases so no big deal.

site/content/notes/Nx.md
```diff
@@ -0,0 +1,225 @@
...
+  - it can be created just for better code organization
+  - ease of re-use can be a positive side effect though
+- library doesn't have to be publishable
+- library doesn't have to be buildable
+  - an app can consume it and build it itself directly
+  - it can be buildable though if we want to enable incremental builds
+- libraries can be nested into sub-folders
+- application can be thought of as a container that budles functionalities implemented in libraries
+
+> We could probably encapsulate some parts of the current tamplate in libs (e.g. component's library, configs etc.). Not only so that in the future we can use them in other templates but also for the sake of separation of concerns, cleanliness, "API"
```
rufuspollock, line 30: This is #todo and should become an issue. This is a very good point.

site/content/notes/Nx.md
```diff
@@ -0,0 +1,225 @@
+# Nx
+
+## Applications and libraries
```
rufuspollock, line 3: This whole document is **excellent** and a super useful summary. 👏👏 It also illustrates the kind of "document/braindump" as learning approach that i think is great.

rufuspollock (approved):
@olayway generally LGTM on the bits i could review (there are 147 changed files to review and most are cosmetic changes!) In general know you are very thorough and looks good.

PR #255, opened 2022-11-08
224 data field type

### LINKED ISSUE #224: Add docs page for `data` contentlayer document field type
Feature added in PR #221

olayway:
Closes #224

## Changes
- small fix to `getPageData` lib function to support async getters (e.g. for external data fetching)
- add basic info about `data` frontmatter field

## Note
We should probably come back to this when we add support for defining custom document types with custom fields, which could then be used as metadata of imported files in getters for filtering and other manipulations (like we do for ecosystem plots). Or at least add a list of fields available on the documents by default (i.e. `sourceFilePath` etc.)


rufuspollock (approved):
Looks good.

PR #280, opened 2022-11-14
Update remark-wiki-link-plus and add test files

### LINKED ISSUE #267: Support capitalized file and directory names

khalilcodes:
### Summary
Resolves #267 

This PR updates the remark-wiki-link-plus package and adds files in `test/` folder.

### Changes

* Latest patch fix in remark-wiki-link-plus `v1.1.1` for case sensitive url paths.
* test files - [test/case-sensitive-wikilinks](https://deploy-preview-280--spectacular-dragon-c1015c.netlify.app/test/case-sensitive-wikilinks) and [test/Test/Page](https://deploy-preview-280--spectacular-dragon-c1015c.netlify.app/test/Test/Page)


olayway:
@khalilcodes there is not much to review here and I think if you wanted me to review the actual changes you could have also created a PR for the plugin itself 😉 ...which again brings this to my mind:

I think it would be much better if we integrated all our remark plugins into our flowershow repo. Any changes we make to them are related to how we want them to work in the template anyway... and it would make both the development and testing much easier, and changes related to one issue could be included in a single PR (e.g. in this case this PR would also include all the changes to the plugin itself, which makes more sense for me as a reviewer). Atm for some reason all the changes to the plugins are just applied without any code review (that is not to say the code in there is bad or sth, but rather that other team members miss on the opportunity to participate in/follow their development process). And it doesn't make any difference for end users of these packages - they will be published separately to npm anyway.

I can't really see any significant reasons why we wouldn't do this tbh 😅.
@rufuspollock

PR #282, opened 2022-11-15
Basic CLI telemetry

### LINKED ISSUE #244: Basic anonymous telemetry on CLI
Normally into google analytics as backend.

## Acceptance

- [x] Defined approach and estimate
- [x] Tracking raw data on runs of `install`, `build` (`build-export`) etc so that we can get counts per week
- [x] info on machine OS
- [x] info on node version
- [ ] (perhaps) error reports (though with no detail) j just that an error happened and command being run

## Tasks

- [x] Research options and estimate time (note we did telemetry in IIRC https://github.com/datopian/data-cli)
  - [x] what library
  - [x] what backend (google analytics?)
- [ ] ...

## Notes

There are two components  in play:
- client: app instrumentation
- backend: data collection/telemetry service

#### Self-hosted

This will require hosting our own backend to collect, analyze and visualize the data.

Most popular tracing/data collection backends:
- [Jaeger](https://www.jaegertracing.io/)
- [Prometheus](https://prometheus.io/)
- [Zipkin](https://zipkin.io/pages/tracers_instrumentation.html)
- [Sentry](https://docs.sentry.io/platforms/node/) (also hosted option)
- [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/)
- ...

#### Hosted

- [Rollbar](https://rollbar.com/)
  - free
    - 5'000 error events/mo
  - $12.5/mo
    - 25'000 error events/mo
- [Sentry](https://docs.sentry.io/platforms/node/)
  - free
    - 5'000 error events/mo
  - $26/mo
    - 50'000 error events/mo
- [bugsnag](https://www.bugsnag.com/) - free for one user or $29/mo
- [Honeybadger](https://www.honeybadger.io/) - free for up to 3 users or $39/mo
- ...

### Client options

- all the tools listed above (apart from Jaeger) have their own instrumentation libraries

### Developers' opinions

Sentry and Rollbar seem to be liked most

https://www.reddit.com/r/laravel/comments/fkv6iy/whats_everyones_preferred_error_tracking_service/
https://www.reddit.com/r/devops/comments/uzdcvz/opinions_on_different_error_monitoring_systems/
https://www.reddit.com/r/rails/comments/ryvi6d/best_ways_to_keep_log_of_and_be_alerted_to_any/

### Datopian data-cli approach

If I understand it correctly, there are two packages used for two different telemetry types:

- raven (old Sentry SDK) - for error tracking: what version of the CLI is being run, on what OS, and what exception was thrown

https://github.com/datopian/data-cli/blob/11a3ab943f58ddd32c6b0d71b3500c9feb762c12/lib/utils/error.js#L11-L31

- universal-analytics - for logged in users to associate their CLI activity with the activity on datahub website, to count/track first-runs of the CLI (along with the OS), to count/track all runs of the CLI (along with the OS and CLI version), to count/track commands being run and with what arguments

https://github.com/datopian/data-cli/blob/11a3ab943f58ddd32c6b0d71b3500c9feb762c12/bin/data.js#L105-L120

also: to track/count successful runs of particular commands (successful first data pushes and other data pushes)
https://github.com/datopian/data-cli/blob/11a3ab943f58ddd32c6b0d71b3500c9feb762c12/bin/data-push.js#L161-L177

### Recommendation

I'd do sth similar do what datopian has implemented in their data-cli. So:

- Just for counting command runs and information about the used CLI version and OS we could also use Google Analytics (universal analytics will be deprecated in Jul next year, so we would use the recommended Google Analytics 4).
- For error tracing, we could use e.g. Sentry or Rollbar starting with a free plan (I don't think we want to host our own backend - it's more time-consuming both in terms of the upfront cost of development and further maintenance) 

### Time estimate

1-3h for GA 4 + 2-4h for error tracking (e.g. with Sentry)

olayway:
Closes #244

### DIFF
--- a/packages/cli/src/bin/cli.js
+++ b/packages/cli/src/bin/cli.js
@@ -1,9 +1,10 @@
 #!/usr/bin/env node
+import os from "os";
 import { createRequire } from "node:module";
 const require = createRequire(import.meta.url);
-import { warn } from "../lib/utils/index.js";
-
-import os from "os";
+import { warn, sendEvent } from "../lib/utils/index.js";
+import { Command } from "commander";
+import inquirer from "inquirer";
 
 if (os.platform() === "win32") {
   warn(
@@ -14,7 +15,22 @@ if (os.platform() === "win32") {
 // TODO check current vs required node version (package.json engines)
 // const requiredNodeVersion = require("../package.json").engines.node;
 
-import { Command } from "commander";
+// ask the user for permission to send anonymous telemetry
+if (process.env.FLOWERSHOW !== "dev") {
+  const { allow } = await inquirer.prompt([
+    {
+      name: "allow",
+      type: "confirm",
+      message:
+        "If you don't mind, we'd like to collect very basic anonymous telemetry that will help us improve this tool. Is that ok?",
+    },
+  ]);
+
+  if (allow) {
+    const [, , cmd, ...args] = process.argv;
+    sendEvent(cmd, args);
+  }
+}
 
 const program = new Command();
 
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -1,5 +1,11 @@
 # flowershow
 
+## 0.0.11
+
+### Patch Changes
+
+- Basic anonymous telemetry (OS, Flowershow version and command being run).
+
 ## 0.0.10
 
 ### Patch Changes
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
-  "name": "@flowershow/cli",
-  "version": "0.0.7",
+  "name": "flowershow",
+  "version": "0.0.10",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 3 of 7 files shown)


packages/cli/src/bin/cli.js
```diff
@@ -14,7 +15,22 @@ if (os.platform() === "win32") {
 // TODO check current vs required node version (package.json engines)
 // const requiredNodeVersion = require("../package.json").engines.node;
 
-import { Command } from "commander";
+// ask the user for permission to send anonymous telemetry
+if (process.env.FLOWERSHOW !== "dev") {
+  const { allow } = await inquirer.prompt([
```
rufuspollock, line 20: Does this prompt every time? or just the first time.

IMO i think for now we could just default this and not prompt. We aren't collecting *any* personal info etc so i don't think there is an issue collecting info and i don't think we need to warn. So in fact i suggest removing for now.

rufuspollock (approved):
LGTM. Only thing would be to remove the prompt to collect info IMO.

PR #294, opened 2022-11-18
[package.json][xs]: forcing update of d3-color

### LINKED ISSUE #200: Investigate high severity vulnerabilities
Investigate and fix 11 high severity vulnerabilities reported by npm.

PhilippeduPreez:
fixes #200 

# Changes
- Used https://www.npmjs.com/package/npm-force-resolutions to force update of package causing problems
- Removes all vulnerabilities as some packages depend on vulnerable versions of d3-color

### DIFF
--- a/packages/template/package-lock.json
+++ b/packages/template/package-lock.json
@@ -7,1335 +7,126 @@
     "": {
       "name": "@flowershow/template",
       "version": "1.0.0",
+      "hasInstallScript": true,
       "dependencies": {
-        "@docsearch/react": "^3.2.0",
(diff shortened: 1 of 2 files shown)


rufuspollock:
@PhilippeduPreez i'd need some detailed explanation of why we need to do this. It seems to add something substantial to the process of installation etc which is a maintenance burden going forward. Is this just to fix one package and what is the issue it is fixing?

PhilippeduPreez:
This is blocked for now https://github.com/flowershow/flowershow/issues/200#issuecomment-1323362421

PR #339, opened 2022-12-02
Blog layout component and blog authors

### LINKED ISSUE #239: Authors attribute for pages with associated author info
Supporting have a database of authors with metadata like fullname and image and use that on pages and posts that have author/authors frontmatter. Nice to have functionality where in absence of a database match you can just use the name as it is ... (?)

### Details

Expected behavior:

```
# in page frontmatter
---
authors: [ rufuspollock]
---

in ... TODO (authors directory?)
TODO
```

### Look

Take inspiration from e.g. nextjs blog

image

### Implementation

suggest following pliny again here in its blog functionality

### Questions

- use `author` or `authors`?
- support single value *and* array or just array

PhilippeduPreez:
## Changes

- added `blog` layout component
- added `authors` type in contentlayer

## Notes
- Currently a default author will be given to all pages, if we only want eg. blog posts to receive author data as props, we could use a conditional in `[[...slug]].jsx` to only do this for blog posts
- closes #239


olayway:
@PhilippeduPreez can you please check why the checks are failing? :)

packages/template/contentlayer.config.js
```diff
@@ -33,6 +33,10 @@ const computedFields = {
     /* eslint no-underscore-dangle: off */
     resolve: (post) => post._raw.flattenedPath,
   },
+  slug: {
+    type: "string",
+    resolve: (post) => post._raw.flattenedPath.replace(/^.+?(\/)/, ""),
+  },
```
olayway, line 39: You don't need to create another computed field just to find a specific author. You can just use his name.

packages/template/contentlayer.config.js
```diff
@@ -65,6 +69,22 @@ const Blog = defineDocumentType(() => ({
...
+  name: "Author",
+  filePathPattern: "authors/**/*.md*",
+  contentType: "mdx",
+  fields: {
+    name: {
+      type: "string",
+      required: true,
+    },
+    avatar: {
+      type: "string",
```
olayway, line 82: This should have a default value. And while I like the one you picked, let's use that boring gray and white icon with the shape of the head 😅

site/content/authors/default.mdx
```diff
@@ -0,0 +1,5 @@
+---
+name: default
+avatar: /assets/images/avatar.png
+type: Author
+---
```
olayway, line 5: You can define default values for contentlayer document type fields, like we do e.g. here `layout: { type: "string", default: "blog" }` 😜

olayway (changes requested):
Please take a look at my comments and let me know if you need some clarification.
Apart from the required changes there are also authors missing in the blogs list.

PhilippeduPreez:
@olayway >there are also authors missing in the blogs list. If you mean the blog index page, I didn't see authors in the examples we drew inspiration from, so only displayed them in the actual blog post

olayway:
After a looong debugging I've managed to make the tests from the `blog.spec` pass. It seems that when there is no `search` property set in the user config blog pages will show the following error, but no idea why only on blog pages... I'll continue working on it tomorrow morning.

image

olayway:
> After a looong debugging I've managed to make the tests from the `blog.spec` pass. It seems that when there is no `search` property set in the user config blog pages will show the following error, but no idea why only on blog pages... I'll continue working on it tomorrow morning.

It seems this is somehow related to `blog` layout and importing `allAuthors` from contentlayer, as replacing the import with just `const allAuthors = []` seems to "fix" the issue. But I have no idea why and how these two things are related...

PR #345, opened 2022-12-07
Tutorial: Customize your site's fonts and colors

### LINKED ISSUE #292: Tutorial: "How to theme your Flowershow website with fonts/colors"
Quick tutorial / howto on modifying your theme using tailwind.config.js

Note: this involves them messing with the `.flowershow` folder so flag that ...

## Job Stories

When i want to theme my site i want to change the default font (i.e. used for navbar, footer, main text) so that ...

When i want to theme my site i want to change the font for headings (and not rest of stuff)

[Low] When i want to them my site i want to use a custom font for just a single section so that ...

### Colors

Want to customize

- background color (in dark and light mode)
- text color (in dark and light mode)
- ...?

## Acceptance

Tutorial doc page on:

- [x] how to customize colors **06/12/2022 https://hackmd.io/6GfGr0TBQBuslEgYVOYviQ**
- [x] How to customize fonts **06/12/2022 https://hackmd.io/6GfGr0TBQBuslEgYVOYviQ**

## Tasks

- [x] tutorial on how to add custom colors
- [x] tutorial on how to add custom fonts
- [ ] Add a commit changing the font for whole site (in a nice way) using a google font **only pushing this after font/color fixes has been made in core**
- [ ] Add a commit changing the key colors (e.g. font, background) **only pushing this after font/color fixes has been made in core**
- [x] update tutorial as per detailed comments in PR #345

## Notes

- Could go into more detail on how to apply these to the whole website instead of just specific pages eg. where you would add these classes in `layout/docs`

PhilippeduPreez:
See https://deploy-preview-345--spectacular-dragon-c1015c.netlify.app/docs/custom-theme for the tutorial

## Changes

- How to add Custom colors and font tutorial

## Notes

- One unrelated minor fix to exclude `meta` from prettier as o/w errors occurring

closes #292

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,3 +7,4 @@ out/
 dist/
 excalidraw/
 notes/
+meta/
--- a/site/content/docs/custom-theme.md
+++ b/site/content/docs/custom-theme.md
@@ -1,0 +1,107 @@
+---
+title: How to modify your theme
+---
+
+## Adding custom fonts to your site
+
+_Adding a google font_
+
+Let's say you want to add a font from fonts.google.com then you would perform the following steps: **I going to add a font, _Kanit_, to my website.**
+
+1. Go to fonts.google.com
+2. Select the font you want
+
+![[kanit.png]]
+
+3. Select the styles you want
+
+![[style.png]]
+
+4. Press `View selected families` at the top right
+5. Copy everything inside the `<style>` tags
+
+![[copystyle.png]]
+
+6. Paste what you copied at the top of the `global.css` file, which can be found in your flowershow directory's style folder - `.flowershow/site/styles/global.css`
+
+```css=
+@import url('https://fonts.googleapis.com/css2?family=Kanit:wght@900&display=swap');
+```
+
+7. Open the file `tailwind.config.js` in `.flowershow` and add the new font by extending the fontFamily property of `theme` as shown below
+
+```javascript
+module.exports = {
+  theme: {
+    extend: {
+      fontFamily: {
+        kanit: ["Kanit"],
+      },
+    },
+  },
+};
+```
+
+Once added, you would then be able to use them in your tailwind classes by prefexing with `font-[name of the font]` eg. `font-kanit` in the above case.
+
+So the following:
+
+```
+# Welcome to my Flowershow site!
+
+<div className="font-kanit">
+All of this text should be my custom font
+</div>
+
+<div>
+All of this text should be the default font
+</div>
+```
+
+Will change from this:
+
+![[flowerbeforefont.png]]
+
+To this!
+
+![[flowerafterfont.png]]
+
+> If you want all your markdown pages to use the new 'kanit' font from above then you can do so by adding it as a tailwind class in the docs layout file found in the layouts folder `flowershow/site/layouts/docs.jsx`
+
+## Adding custom colors
+
+To add custom colors in your site, extend the colors property for `theme` in `tailwind.config.js` as shown below
+
+```javascript
+module.exports = {
+  theme: {
+    extend: {
+      colors: {
+        primary: "#1e4558",
+        secondary: "#d52027",
+        "theme-yellow": "#ffb049",
+      },
+    },
+  },
+};
+```
+
+So the following:
+
+```
+<div className="bg-secondary">
+This has the seconday color as background
+</div>
+
+<div className="bg-theme-yellow text-primary">
+This has theme-yellow as backgound and primary as text
+</div>
+```
+
+Will result in this:
+
+![[changecolor.png]]
+
+After adding the colors, you can use them in your tailwind classes for text, background, etc. For example taking the above values you can set the background color to your secondary color with `bg-secondary` or set the text color to your primary color with `text-primary`
+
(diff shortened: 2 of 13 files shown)


site/content/docs/custom-theme.md
```diff
@@ -0,0 +1,110 @@
+---
```
rufuspollock, line 1: Structural comments on overall piece:

- I would have a section, probably at the start about tailwind and its role in Flowershow
  - explain that Flowershow uses tailwindcss. Point people to our reference docs on this (do we have them) and/or tailwindcss docs.
  - Explain that: thus customization of fonts and colors is going to be done via tailwindcss and especially its config file (against link docs)
  - Explain where to find this file (try and make this generic e.g. file is called `tailwind.config.js` and located in the directory where your flowershow app is installed, usually a `.flowershow` subfolder of wherever you ran `npx flowershow install`.
- then have section on customizing fonts
  -  explain what this means e.g. we are going to change heading font or body font across whole site
    - maybe note that if they just want to change one section that should be done using html/tailwindcss (TODO: a future tutorial perhaps ...)
  - then proceed as normal
- ditto for customizing colors

site/content/docs/custom-theme.md
```diff
@@ -0,0 +1,110 @@
...
+description: Learn how to add custom fonts and colors to your site
+type: Blog
+created: 2022-12-07
+---
+
+## Adding custom fonts to your site
+
+_Adding a google font_
+
+Let's say you want to add a font from fonts.google.com then you would perform the following steps: **I going to add a font, _Kanit_, to my website.**
```
rufuspollock, line 12: I think we could assume they know how to find a font e.g. on google fonts.

site/content/docs/custom-theme.md
```diff
@@ -0,0 +1,110 @@
+---
+title: How to modify your theme
+description: Learn how to add custom fonts and colors to your site
+type: Blog
+created: 2022-12-07
+---
+
+## Adding custom fonts to your site
+
+_Adding a google font_
```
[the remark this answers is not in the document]
rufuspollock: I think the outline of this section is:

- Motivation (not there yet)
- Get a font and make it available (can be short on how to find it and focus on adding to global.css)
- Use that font in tailwind.config.js (we can focus on what we are customizing and provide links to taildwind docs for more)

site/content/docs/custom-theme.md
```diff
@@ -0,0 +1,110 @@
...
+    extend: {
+      fontFamily: {
+        kanit: ["Kanit"],
+      },
+    },
+  },
+};
+```
+
+Once added, you would then be able to use them in your tailwind classes by prefexing with `font-[name of the font]` eg. `font-kanit` in the above case.
```
rufuspollock, line 48: I actually think this use-case is quite rare (i.e. i want to have a different font for just one area) and also covers an additional point which is using html and tailwind css in markdown files (a good topic for another short how to).

I think focus should be on changing font throughout site e.g. changing the serif or sans font which changes font for all headings.

site/content/docs/custom-theme.md
```diff
@@ -0,0 +1,110 @@
...
+
+Will change from this:
+
+![[flowerbeforefont.png]]
+
+To this!
+
+![[flowerafterfont.png]]
+
+> If you want all your markdown pages to use the new 'kanit' font from above then you can do so by adding it as a tailwind class in the docs layout file found in the layouts folder `flowershow/site/layouts/docs.jsx`
```
rufuspollock, line 72: I think the better approach is to change fonts in tailwind.config.js - rather than this approach.

In fact, we should check that we are using tailwind.config.js for this.

site/content/docs/custom-theme.md
```diff
@@ -0,0 +1,110 @@
...
+</div>
+```
+
+Will result in this:
+
+![[changecolor.png]]
+
+After adding the colors, you can use them in your tailwind classes for text, background, etc. For example taking the above values you can set the background color to your secondary color with `bg-secondary` or set the text color to your primary color with `text-primary`
+
+> If you want all your markdown pages to use the newly added colors then you can do so by adding it as a tailwind class in the docs layout file found in the layouts folder `flowershow/site/layouts/docs.jsx`
```
rufuspollock, line 110: Again i think we want layouts to *already* use tailwind primary etc rather than having to go modify layouts (among other things may be several layouts).

/cc @khalilcodes i assume we have changed them to start using primary, secondary etc (?)

rufuspollock (changes requested):
Extensive comments on structure and content. Hope they are useful.

.prettierignore
```diff
@@ -7,3 +7,4 @@ out/
 dist/
 excalidraw/
 notes/
+meta/
```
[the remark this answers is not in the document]
PhilippeduPreez: This was from an error occurring with husky when a changes are directly push to main as I understand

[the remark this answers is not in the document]
rufuspollock: OK. Good to make a note of that in the PR e.g. "One unrelated minor fix to exclude `meta` from prettier as o/w errors occurring"

PhilippeduPreez:
Opened new PR for this #391

Issue #36, opened 2022-07-14
Link previews on hover


olayway:
Distinguish external and internal link previews. For now focus on internal links.

Note we already implemented this in web3.lifeitself.org so i think a question of porting this over - see https://github.com/life-itself/web3/issues/150 and PR https://github.com/life-itself/web3/pull/162

## Acceptance

- [ ] When i hover over an (internal) link i get a preview of the contents (e.g. first X words) on that page
  - [ ] (also maybe the featured image if there is one?)

Example from https://web3.lifeitself.org/guide

<img width="691" alt="image" src="https://user-images.githubusercontent.com/180658/199662363-693ad2f5-9b32-4d09-a310-2186ee51bbe2.png">

## Tasks

- [x] Research and estimate this **✅ 2022-12-13 Porting over from web3**
- [x] Create docs with guide to toggle off **✅ 2022-12-14**
- [x] Create the tooltip component **✅ 2022-12-14**
  - [x] modify to parse the content body **using https://github.com/markedjs/marked**
  - [x] modify to use frontmatter image in preview if available
- [x] Add option to disable in `config.js` **✅ 2022-12-14**

khalilcodes:
@olayway do we want this to be optional in config ?

rufuspollock:
@khalilcodes great work.

> 
@olayway do we want this to be optional in config ?

IMO we could just make this default behaviour for now and not worry about making it optional. But, if that is very cheap to do, then do it.

> modify to parse the content body using [markedjs/marked](https://github.com/markedjs/marked)

Out of interest why are we using marked when we already have remark? does that mean we now have a bundle dependency on two markdown parsers?

@khalilcodes btw where's the estimate for this - i don't see any time estimate.

<img width="327" alt="image" src="https://user-images.githubusercontent.com/180658/207704563-f2ee4747-70ae-42c2-ba00-3af4eca3950f.png">

khalilcodes:
@rufuspollock 

> Out of interest why are we using marked when we already have remark? does that mean we now have a bundle dependency on two markdown parsers?

we would have to add remark-parse and unified and extract the required content like we do in web3. I used marked instead cause it's client side, but can be replaced.

rufuspollock:
@khalilcodes is this done? can we update acceptance and close?

rufuspollock:
@khalilcodes is ready to go and is opening a PR.

rufuspollock:
FIXED. Copying from #382 

### Summary

[...]

PR #360, opened 2022-12-14
[content][s]: documentation for link preview tooltips

khalilcodes:
### Summary
Tasks from #36 

This PR adds documentation about link preview tooltips on hover and how to disable them in `config.js`

### Changes
* Add docs page at [docs/link-previews](docs/link-previews)
* Add screenshot images

### DIFF
--- a/site/content/assets/images/link-preview-with-image.jpg
+++ b/site/content/assets/images/link-preview-with-image.jpg
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:fa49ca726b38d4471a989b952a0877d27dd6ffacfed984228fb7e35b883c9751
+size 39044
--- a/site/content/assets/images/link-preview.jpg
+++ b/site/content/assets/images/link-preview.jpg
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:0b5bc151a81d624852bd5646069413c6c4cdaed2ee1dc2d976ff950bfd03a02d
+size 46802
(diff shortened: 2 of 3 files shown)


khalilcodes:
@olayway I added some documentation first before opening the PR of the actual feature. I have kept it to be enabled by default and added in docs on how to disable them if needed. Let me know if this looks right. :)

olayway (approved):
Looks good! 👏

One minor thing: do you think it would be possible to add functionality to disable previews OF selected pages? E.g. here the preview content for the roadmap page is not really useful and it would be nice to not show it for this particular page.
I imagine it could be done in the config file using a list of glob patterns or/and in frontmatter of selected pages. But if it's not easy to implement I'd add it in the next iteration. Up to you 🙂

rufuspollock:
> One minor thing: do you think it would be possible to add functionality to disable previews OF selected pages? For some pages, it just doesn't make much sense. E.g. here the preview content of the roadmap page is not really useful and it would be nice to not show it for this particular page. (Or for example for some index pages or the home page). I imagine it could be done in the config file using a list of glob patterns or/and in frontmatter of selected pages. But if it's not easy to implement I'd add it in the next iteration. Up to you 🙂

Great suggestion Ola, and i can clarify (for me as product owner) this is **not** a need - for now. Let's keep things super-simple and add options like this later based on experience 😄

rufuspollock:
> Also in the future, I'd maybe use some other page as an example of a preview with an image. Roadmap's preview content is not a particularly good advertisement of the preview feature's usefulness 😅

Very good point.

Issue #351, opened 2022-12-10
Use named fonts and colors in default theme (and style guide for developers re this)


rufuspollock:
When developing or customizing the default theme I want to to use standard named fonts and colors so that it is easier to change the theme coherently going forward

We currently tend to implement ad-hoc styling e.g. here is a snippet of Avatar.jsx

```
<div className="ml-3">
  <p className="text-sm font-medium text-black dark:text-white">
    {name}
  </p>
</div>
```

Should be like:

```
<div className="ml-3">
  <p className="text-sm font-medium text-primary dark:text-primary-dark">
    {name}
  </p>
</div>
```

We should be using standard named colors from tailwind.config.js so that

- it is easy for us to maintain consistency
- it is easy to for a user to systematically change colors across the whole site 

Our default theme does its styling for fonts using font-sans and font-serif. For colors people just add colors ad-hoc as they want. This means that making consistent changes to the theme either for developers or users meaning getting into the templates and components themselves which is painful and error-prone.

## Acceptance

- [x] using named semantic fonts for key font areas **✅ Done in PR https://github.com/flowershow/flowershow/pull/361**
- [x] using named semantic colors for key color areas **✅ Done in PR https://github.com/flowershow/flowershow/pull/362**
- [ ] style guide for this **✅2022-12-20 https://flowershow.app/docs/contributors#styling-and-tailwind**

## Tasks

- [x] Analyse how we implement in config
- [x] Add semantic items **✅ 2022-12-15**
- [x] Refactor existing templates and components to use these semantic labels **✅ 2022-12-15**
- [x] Add a note to developer docs to this (style guide?) **✅2022-12-20 https://flowershow.app/docs/contributors#styling-and-tailwind**

### Research

See https://tailwindcss.com/docs/font-family#customizing-your-theme and https://tailwindcss.com/docs/customizing-colors#using-the-default-colors

❗ complication: we use typography to do our default styling. However, we can customize key parts relatively easily https://tailwindcss.com/docs/typography-plugin#element-modifiers

eg. for headings we can do (in templates)

```
<article class="prose prose-img:rounded-xl prose-headings:underline prose-a:color-primary">
  {{ markdown }}
</article>
```

### Design

What areas do we want to customize by default? What's the KISS answer?

- fonts:
  - default font: defaults to font-sans though you can change this. See https://tailwindcss.com/docs/font-family#customizing-the-default-font
[...]

PR #361, opened 2022-12-15
Default fonts in tailwind

khalilcodes:
### Summary
Tasks from #351 

This PR adds tailwind's default fonts and a heading font for easy configuration and use in layouts or components.

### Changes
* Overwrite `tailwind.config.js` with tailwind's default fonts
  * Add a custom `font-headings` property for headings
* Use semantic heading fonts in layouts

### DIFF
--- a/packages/template/tailwind.config.js
+++ b/packages/template/tailwind.config.js
@@ -1,3 +1,5 @@
+const defaultTheme = require("tailwindcss/defaultTheme");
+
 module.exports = {
   content: [
     "./pages/**/*.{js,ts,jsx,tsx}",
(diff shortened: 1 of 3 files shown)


packages/template/tailwind.config.js
```diff
@@ -12,6 +14,12 @@ module.exports = {
       maxWidth: {
         "8xl": "88rem",
       },
+      fontFamily: {
+        sans: ["ui-sans-serif", ...defaultTheme.fontFamily.sans],
+        serif: ["ui-serif", ...defaultTheme.fontFamily.serif],
+        mono: ["ui-monospace", ...defaultTheme.fontFamily.mono],
+        headings: ["-apple-system", ...defaultTheme.fontFamily.sans],
```
rufuspollock, line 21: why do we have to `-apple-system` - can't we just run with default sans?

rufuspollock (approved):
Seems ok - what about headings in the flowershow.app site theme? do we need to switch those too?

PR #362, opened 2022-12-16
[tailwind][s]: named theme colors

khalilcodes:
### Summary
Tasks from #351 

This PR adds named theme colors in tailwind's config for customizing background et al.

### Changes
* Add named color options in `tailwind.config.js` - *primary, secondary and background*
* refactor components and layouts to use semantic colors

### DIFF
--- a/packages/template/components/Avatar.jsx
+++ b/packages/template/components/Avatar.jsx
@@ -11,7 +11,7 @@ export function Avatar({ name, img, href }) {
           />
         </div>
         <div className="ml-3">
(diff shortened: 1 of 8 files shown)


rufuspollock:
@khalilcodes this looks good - one question: what about the prose coloring - don't we need to do that too and in some special way?

rufuspollock:
@khalilcodes also can you refer to relevant issue in relevant commits (not just in the PR) e.g. just have `[....,#351][s]: ...`

PR #373, opened 2022-12-28
Allow installing as non-admin on Windows

aabounegm:
Fixes a minor typo and passes the `type` parameter to [`fs.symlinkSync`](https://nodejs.org/api/fs.html#fssymlinktarget-path-type-callback) (which is used only on Windows and ignored on others) to remove the need to run as admin.

### DIFF
--- a/packages/template/package.json
+++ b/packages/template/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
-    "postbuild": "cross-env NODE_OPTIONS='--experimental-json-modules --experimental-modules' node ./scripts/search.mjs",
+    "postbuild": "cross-env NODE_OPTIONS=\"--experimental-json-modules --experimental-modules\" node ./scripts/search.mjs",
     "start": "next start",
     "export": "npm run build && next export"
   },
--- a/packages/cli/src/bin/cli.ts
+++ b/packages/cli/src/bin/cli.ts
@@ -14,7 +14,7 @@ const { version: node, platform, argv } = process;
 
 if (platform === "win32") {
   warn(
-    "This may not work as expected. You're trying to run Flowreshow CLI on Windows, which is not thoroughly tested. Please submit an issue if you encounter any problems: https://github.com/flowershow/flowershow/issues"
+    "This may not work as expected. You're trying to run Flowershow CLI on Windows, which is not thoroughly tested. Please submit an issue if you encounter any problems: https://github.com/flowershow/flowershow/issues"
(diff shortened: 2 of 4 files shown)


aabounegm:
Do I need to add a changeset to the PR? I do not see it in the instructions in the README, so I hope it's not blocking this PR.

olayway:
@aabounegm thank you so much for submitting this PR 🌼! It looks good to me.
@PhilippeduPreez can you please quickly test if it works? If so, I'll update some related parts of our docs and we can merge!

@aabounegm As for the changesets file - no, I don't think we need to add one for such a minor change. For future reference, here is the docs page with some helpful information for contributors - https://flowershow.app/docs/contributors. If anything is not clear or some information you're looking for is missing, please do let us know! Thanks again!

aabounegm:
Thanks for the positive comment 😁. I did check the page, but I thought this fix was too small to open an issue for it before, so I got worried when the PR took a few days to be reviewed 😅. I tried to follow your commit messages style, though. Hope I didn't do it badly 😄.

packages/template/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "next dev",
     "build": "next build",
-    "postbuild": "cross-env NODE_OPTIONS='--experimental-json-modules --experimental-modules' node ./scripts/search.mjs",
+    "postbuild": "cross-env NODE_OPTIONS=\"--experimental-json-modules --experimental-modules\" node ./scripts/search.mjs",
```
aabounegm, line 8: Sorry, I took the freedom to fix this one as well before the review (it doesn't directly relate to the PR title, but it still fixes a Windows-only problem) 😅. Apparently, Windows (or CMD maybe?) doesn't like the single quotes, but the double quotes should work on most platforms. But also, I can't help but wonder why not use the simpler form (that doesn't depend on `cross-env`):
```suggestion
    "postbuild": "node --experimental-json-modules --experimental-modules ./scripts/search.mjs",
```
?

aabounegm: This was causing another silent fail on Windows with `pnpx flowershow export`. I was wondering why it generated the `.next` output but not the static one, and figured out that this is the root cause.
This was (hopefully) the last issue stopping me from adopting Flowershow for my PKM (there are a few other minor issues, but nothing critical).

rufuspollock:
@PhilippeduPreez any thoughts?

rufuspollock:
@aabounegm thank you so much for the contribution! We'll have a shout-out to you in our next release update!

PR #387, opened 2023-01-22
Template switch from JS to TS

### LINKED ISSUE #13: Use typescript
Shall we use typescript? Pros and cons please ... **UPDATE: we will trial it ...**

## Acceptance

* [x] Made a choice about trialling
* [x] Do basic setup (what is required?)
* [x] Start using
* [x] Summarize experience and whether we go 100%

## Tasks

- [x] migration analysis
- [x] switch from JS to TS in CLI package **done #300**
- [ ] switch from JS to TS in the template package

## Analysis

Since TS is a superset of JS, migration to 100% TS codebase can be gradual. Also, Nx has native TS support so all the workspaces and projects in our repo will already include basic ts.config.json files and should already support TS out of the box.

### Suggested approach

* Add support for TS to the existing codebase
  - ts compiler configurations,
  - add build step for the CLI,
  - start creating `d.ts` declaration files to enable type-checking of the existing JS code by the TS compiler (the additional benefit at this point is auto-completion support in code editors)
  - write all NEW files in TS
* Slowly start refactoring existing JS files to TS.

#### Making template package work with `tsconfig.json` file instead of our original `jsconfig.json`

**TL:DR:**
`npm run dev` --> Next.js sees the `tsconfig.json` file and adds `jsx: preserve` compiler option (or overrides it) --> Contentlayer's esbuild picks up the changed `tsconfig.json` file --> JSX in md files is not converted to JS --> errors!

Problem: contentlayer is not converting JSX to JS when used with Next.js + `tsconfig.json` file.

**Context:** When Next.js sees the `tsconfig.json` file in the project (template), it overrides some of the values that are required to make Next.js work as expected and they don't have the option to opt out of it. One of these properties is `jsx`, which by default is set to `react`, but Next.js will change it to `preserve`.

> react: Emit .js files with JSX changed to the equivalent React.createElement calls
> preserve: Emit .jsx files with the JSX unchanged

Next.js does it because they claim they have better way of converting JSX to JS, so they don't want the ts compiler to do this. However, the same ts.config.js file will be picked up by Contentlayer's esbuild and so it will not convert JSX to JS with React elements.

olayway:
Closes #13

## Changes
- adds separate `tsconfig.mdx.json` file passed directly to contentlayer's `esbuild` (see `contentlayer.config.js`)
- renames template files from `.js` to `.ts`
- renames two files in `core` package to `ts` and corrects the return type of `SearchProvider` (+ version bump)

## TODOs
- [ ] publish updated version of `core` to npm


khalilcodes (approved):
LGTM 👍 I think let's merge this one first, then I can make the changes in Toc PR accordingly.

PR #392, opened 2023-01-23
Fix RHS Table of Contents and headings

### LINKED ISSUE #284: RHS ToC some headings don't show up
Comment from @rufuspollock in https://github.com/flowershow/flowershow/issues/95#issuecomment-1301209125_
   
> @khalilcodes when closing a substantive issue like this can we get a bit more info e.g. a screenshot and/or link to example pages - that is really useful when reviewing the issue.
>
> also wanted to flag what seems like potentially buggy behaviour on e.g. this page: https://flowershow.app/meta/publishing-design. Here a whole set of headings don't show up in ToC on RHS (my guess is because nested within a h1). Questions:
> 
> - I would assume h1 other than very first "title" h1 would show up in ToC (though i'm not totally certain about this or how complex to do0
> - even if h1 does not show up i would imagine the h2/h3 etc underneath it would show up ...
>
> image   

# Acceptance

- [ ] RHS ToC working on pages
- [ ] Missing headings show up

# Tasks
- [x] Move all ToC functions to core
- [x] Headings for h1, h2 and h3
- [x] Page Title heading not in ToC.
  - This is setup to check for fronmatter title or h1 heading as first row (eg. `# Page Heading` and only if frontmatter title doesn't exist)

khalilcodes:
Closes #284 
### Summary

This PR fixes the RHS table of contents not rendering and also some of the headings that were not in the ToC list.

### Changes
* Move ToC functions from template's `_app.jsx` to `Layout.tsx` in core
* Modify logic for `collectHeadings` and `useTableOfContents` hook.
  * The headings now support 3 levels ie. h1, h2 & h3
  * All level headings displayed in list irrespective of parent (like `h1 --> h2`). See [meta/publishing-design](https://deploy-preview-392--spectacular-dragon-c1015c.netlify.app/meta/publishing-design) as an example.


packages/core/src/utils/collectHeadings.ts
```diff
@@ -0,0 +1,67 @@
...
+    const { id, innerText: title, tagName: level } = node;
+
+    if (!(id && title)) {
+      return;
+    }
+
+    if (level === "H1" && (headerTitle ?? !pageHeading)) {
+      sections.push({ id, title, level, children: [] });
+    }
+
```
olayway, line 24: Can you please explain what you're trying to achieve here? Especially on line 21.

olayway (commented):
@khalilcodes can you please add a comment on the general approach you're trying to implement for collecting ToC headings (in `collectHeadings.ts`)? Can you write a few sentences explaining what and why you want to include in the ToC? E.g. this is what happens if you add h2 headings below the first h1 in the `meta/publishing-design` page you shared:

image

So, the question is what do you want to do in case of multiple h1 headings? Do you want to support it and how? Does including h1 tags but without only the first one even makes sense from the perspective of the person reading the page?

packages/core/src/utils/collectHeadings.ts
khalilcodes: In the issue:
> I would assume h1 other than very first "title" h1 would show up in ToC 

Here I'm checking if there's a title in the frontmatter. If there is one, then don't remove the h1 heading from ToC  eg. if markdown starts with `# Some Heading`. If no frontmatter title is set, then h1 is considered as page's title and we remove the h1. 

The only caveat with this approach is that the frontmatter title should be inside a header element (that's how our docs and blog layouts are set up right now).

packages/core/src/utils/collectHeadings.ts
```diff
@@ -0,0 +1,67 @@
+// ToC: get the html nodelist for headings
```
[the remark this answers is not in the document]
khalilcodes: I had that as a seperate commit and then I pushed the last one for homepage and honestly don't know what happened there 🤷

packages/core/src/utils/collectHeadings.ts
khalilcodes: Also as in the shared image above
> even if h1 does not show up i would imagine the h2/h3 etc underneath it would show up ...

Maybe I understood this wrong. What I tried here was to show h2/h3 even if they didn't have a parent. Let me know if this is not the desired outcome and I'll change that :)

packages/core/src/utils/collectHeadings.ts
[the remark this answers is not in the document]
olayway: No problem :)

packages/core/src/utils/collectHeadings.ts
olayway: > They show up as seen above right ?
But I don't think that's a correct representation of the page structure. Take a look at the screenshot once again. You have "Some heading lvl 2" (h2 tag) with the same indentation as "User experience" (h1 tag) 😛

olayway: > I didn't understand the 2nd option, aren't 2 and 3 the same thing ?

Only for pages with multiple h1 tags.

olayway: > Yeah, I think option 3 has better maintainability. I will make that change.

I think so too.

olayway:
@khalilcodes one more thing - changesets 🙂

packages/core/src/utils/collectHeadings.ts
khalilcodes: But now I'm facing another issue with the above ie., what happens in pages where there are only h2 headings for example like we have in our /docs page. In this scenario, the ToC won't render.

Looking at tailwind's [syntax](https://tailwindui.com/templates/syntax) theme, they are only displaying h2 and h3. if we put an h3 without an h2 before it, an error is thrown like below.

Screen Shot 2023-01-23 at 10 15 58 PM

Should we not render h2 if there's no preceding h1 and same for h3  ?

olayway: @khalilcodes I think let's just not remove anything from the ToC. Even we have some pages with multiple h1 tags and I used to write my notes using multiple h1 tags as well. I have also googled around a bit, and while this may be a good idea to only have one h1, there's nothing wrong with having multiple. It all depends on what content you're writing and probably also how many levels of different sections you need, and you may need many, so you'd probably could also use this h1 to additionaly divide your content. The other benefit of not removing h1 tags is that if they do have some text, like e.g. intro to the page, if you remove the first h1 you can't use the right hand side ToC to quickly jump to the very first heading.

olayway: And in this case the ToC clearly reflects all the different sections on your page, nothing less nothing more. Probably even for screen readers this might be a better approach.

PR #401, opened 2023-01-25
Add typed animations in Hero section

### LINKED ISSUE #307: Main page title variations (with typed.js?)
Currently main title is focused on "Obsidian" just because we had to pick one item at start:

> Publish your Obsidian notes, beautifully.

Suggest we could vary this e.g.:

- Publish your markdown notes, beautifully.
- Publish your [markdown] docs, beautifully.
- Publish your digital garden, beautifully.
- Publish your second brain, beautifully.

Other thoughts

- do we also try out varying "beautifully". Key aspects are: beautiful/elegant, easy, fast e.g. "Publish your digital garden easily and elegantly"

## Acceptance

- [ ] agreed list of variations
- [ ] cycle through these on the main page

## Tasks

- [ ] create list of options
- [ ] implement - suggest we use: https://github.com/mattboldt/typed.js/

khalilcodes:
Closes #307 
### Summary

This PR adds a typed text animation for the Hero title

### Changes
* Add typed.js package in template
* Change Hero title with typed animation text
* update pnpm lock file

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -243,6 +243,7 @@ importers:
       remark-smartypants: ^2.0.0
       remark-toc: ^8.0.1
       tailwindcss: ^3.1.8
+      typed.js: ^2.0.12
     dependencies:
       "@flowershow/core": link:../core
       "@flowershow/remark-callouts": link:../remark-callouts
@@ -271,6 +272,7 @@ importers:
       remark-math: 5.1.1
       remark-smartypants: 2.0.0
       remark-toc: 8.0.1
+      typed.js: 2.0.12
     devDependencies:
       autoprefixer: 10.4.13_postcss@8.4.20
       cross-env: 7.0.3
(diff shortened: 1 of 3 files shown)


khalilcodes:
@olayway  what do you think, should we have typed or faded animations here ?

olayway (commented):
@khalilcodes One thing I'm not sure about is adding `typed.js` as a dependency of the template to use it in a custom component. I'd probably do this if our template and our Flowershow website were separate things, in a way a power user would customize his Flowershow website. But in this case, all users would install it whether they use it or not.

I think maybe moving the `Hero` component and `typed.js` dependency to `@flowershow/core` would be a better idea. In the future, we could probably even refactor this component and make it customizable so that users can use it on their own home pages as well.

olayway:
For `BlogsList` component (also part of `@flowershow/core`) I had to pass it to MDX in `[...slug]` as importing it directly in MDX caused some errors, but now I'm trying it out again and it seems to be working correctly. E.g. in this case you'd do:

```
import { Hero } from "@flowershow/core"

<Hero />
```

olayway:
> @olayway what do you think, should we have typed or faded animations here?

@khalilcodes I think typed are fine but atm depending on the length of the typed phrase hero changes height which doesn't look good. Also on smaller screens it doesn't look nice. But I guess these are just some minor CSS tweaks to do.

pnpm-lock.yaml
```diff
@@ -243,6 +243,7 @@ importers:
       remark-smartypants: ^2.0.0
       remark-toc: ^8.0.1
       tailwindcss: ^3.1.8
+      typed.js: ^2.0.12
```
rufuspollock, line 246: Hmmm, i get what @olayway is saying. We really don't want to have to add this to core to have this for the main site. I guess we are getting to point where we want website to be its own app properly. This is something worth flagging IMO. For now we live with it but we definitely want to flag for ourselves as we move forward.

khalilcodes:
@olayway based on @rufuspollock comments, did not move to core and fixed the height and responsive issue

Issue #94, opened 2022-08-16
Full site Table of Contents


rufuspollock:

StefanoCecere:
for me this is the only feature needed to jump to flowerapp for my sites

rufuspollock:
@StefanoCecere this will be worked on soon! (i would guess early january at this point given the holidays).

StefanoCecere:
i just created a test Tailwind site.. but it seems that the left sidebar component is a commercial product!

obstschale:
I was just testing flowershow and I do miss that feature. I would love to see the LHS table of content to auto-populate by all notes on my site. Manually maintaining navigation is a cumbersome task

olayway:
@obstschale thank you for your comment!

@rufuspollock let's maybe add this issue to our next sprint.

rufuspollock:
This is next up now! We plan to work on this in the next sprint or so.

olayway:
@rufuspollock here are some notes on implementation. All in all, this seems like a pretty easy feature to implement, at least in the first iteration where we just display all the pages and don't group them or allow for other customizations.

https://hackmd.io/seZsy4lPQQWQ5Xu4JAFbtg

rufuspollock:
@olayway i moved the notes inline in the description.

i think your 1st iteration sounds good so let's start implementing.

obstschale:
Awesome. Thx for this MVP. It simply works. 

BTW: I like your iteration-approach. Looking forward to functionality, like grouping. 👍

StefanoCecere:
the way Nextra uses to config the menus is really nice!
https://nextra.site/docs/docs-theme/page-configuration

olayway:
@StefanoCecere I love it too! We'll work on making our content a little bit more beautiful in the next iteration and I'll add the link to Nextra's docs as an inspiration ;)

rufuspollock:
@olayway can we update this issue with acceptance and updates on the acceptance items. Also could you link to the documentation for this feature (at the top of the description preferably with a note like "progress is being made look here" type of thing.

homostellaris:
Hello 👋 hope you don't mind me jumping in here but would love to see mobile support 🙏 is there any way that I can help with this?

rufuspollock:
@homostellaris sorry for delayed reply - we'd love your help in improving this if you'd like to take a look.

homostellaris:
> @homostellaris sorry for delayed reply - we'd love your help in improving this if you'd like to take a look.

[...]

PR #403, opened 2023-01-26
LHS sidebar with site-wide table of contents v.1

olayway:
Related: #94

## Changes
- adds a first iteration of LHS sidebar with site-wide ToC
  - pages pulled from `search.json` file (for now; will rethink in next iterations when we want to allow for some more customization) 
  - all pages included and sorted alphabetically (based on title or slug)
  - global config option `showSidebar` to enable/disable it on all pages
  - per-page config option `showSidebar` to enable/disable on single pages
- refactors core's `Layout` component into several child components
- docs page at `content/docs/site-wide-toc`
- simple e2e tests

## Notes
- sidebar is only visible on large+ screens atm; I'll add mobile support in the next iteration

## TODOs
- [ ] publish core to npm


olayway:
@khalilcodes I've just added a doc page

olayway:
@khalilcodes I've just added basic e2e tests as well

packages/core/src/ui/Layout/Layout.tsx
```diff
@@ -60,166 +84,59 @@ export const Layout: React.FC<Props> = ({
...
-        <div className="relative mx-auto">
-          <main className="flex-auto">
+        {/* NAVBAR */}
+        <div
+          className={clsx(
+            "sticky top-0 z-50 w-full",
+            isScrolled
+              ? "dark:bg-background-dark/95 bg-background/95 backdrop-blur [@supports(backdrop-filter:blur(0))]:dark:bg-background-dark/75"
+              : "dark:bg-background-dark bg-background"
+          )}
```
khalilcodes, line 94: why did we move this from nav component to here ?

site/content/docs/site-wide-toc.md
```diff
@@ -0,0 +1,37 @@
...
+};
+```
+
+### For a single page
+
+You can also enable/disable the sidebar for specific pages by setting `showSidebar` to `true` or `false` in the page's frontmatter like below.
+
+```md
+---
+showToc: false
```
khalilcodes, line 29: showSidebar*

khalilcodes:
@olayway awesome work on separating the components out, I like it. 👏

packages/core/src/ui/Layout/Layout.tsx
olayway: Since now the Layout component controls the positioning of all the main blocks it's composed of,  I've moved paddings and width classes out of Nav (and out of the other child components it's now composed of), so that we can easily adjust e.g. width of the whole site content (for which you need to adjust paddings, margins or width of the main Layout blocks - in a few different places - e.g. Navbar and content below it). And having e.g. max width applied to these components only in the Layout makes it easier, as it's easy to find. So now e.g. Nav component is everything that's inside, i.e. links, search field etc. But we want to apply the background on the whole container. That's why it's moved to Layout. Also, I just think telling if the site has scrolled or not is rather a job of the whole Layout and in this case, it can be used to style also other components, not only the Nav. Or e.g. based on its value we could display an additional component - a button to scroll to the top, or sth like that :) Hope that makes sense 😅

site/content/docs/site-wide-toc.md
olayway: Thank you!

packages/core/src/ui/Layout/Layout.tsx
khalilcodes: Sounds like a plan 😉

khalilcodes: Now that there are some more paddings in the layout the typed animation in the Hero will lap again, but we can fix that.

khalilcodes (approved):

PR #406, opened 2023-01-30
Page comments / discussions

### LINKED ISSUE #163: Page comments / discussions
When publishing a page I want to allow visitors to post comments so that I get feedback

* want to be able to bulk enable/disable e.g. in settings
* enable/disable page by page

Implementation: suggest we build on (and contribute to) pliny, see https://github.com/timlrx/pliny/tree/main/packages/pliny/src/comments

## Acceptance

* [x] Commenting support on posts and pages **✅ 2023-02-01 (we are using discussions "Announcements") merged in commit ...**
- [x] Enable page by page in frontmatter:
```md
---
showComments: true 
---
```
- [x] Bulk enable in config:
```js
const config = {
  comments: {
    pages: ["blog"] // can add more in array eg. docs
  }
}
```
Screen Shot 2023-01-30 at 1 35 11 PM

## Tasks

Research / Design

**✅ 2023-01-25 Using pliny approach but with imported components from providers**

- [x] how does it work roughly
- [x] what do you think of it? (tech and design wise) i.e. pliny implementation
- [x] what is recommended approach in terms of approach and implementation
	- [x] should we follow this approach or look at something else
	- [x] what are the key tasks
- [x] how long will it take guesstimate? (for MVP and polished)

Implementation 

**✅  2023-01-31**

- [x] Create comments components in core for [giscus](https://giscus.app/), [utterances](https://utteranc.es/) and [disqus](https://disqus.com)
 - [x] Add and use [giscus react](https://github.com/giscus/giscus-component) and [disqus-react](https://github.com/disqus/disqus-react) packages in core 
- [x] Add configs in frontmatter and config.js
  * configured to set page comments through frontmatter, or specified folders in comments' config. If both are not set then defaults to show on all pages. Frontmatter value takes precedence.
- [x] Add documentation @ [docs/page-comments](https:/flowershow.app/docs/page-comments)
- [x] publish core package to npm
- [x] Add giscus env variables in netlify
- [x] install giscus app in flowershow repo

## Notes

> 1. how does it work roughly
* provider components are setup in let's say `core/ui/Comments` and rendered based on config values - `GiscusConfig | DisqusConfig | UtterancesConfig`.
* cors setup in `next.config.js` and env variables
* Component is imported in either the whole layout or Blog's layout file if we only want them for posts.

> 2. what do you think of it? (tech and design wise) i.e. pliny implementation
* This works really well for multiple providers and also scalable for adding more. If we are not using more than one provider, then it's probably a single component imported in our `Layout.tsx` which is rendered based on config and pageProps values.

> 3. what is recommended approach in terms of approach and implementation	
> should we follow this approach or look at something else
	 
If we are looking to use a single provider but are not sure about adding more in future, then this approach would be the route to take. My thoughts are since this is a plug it in kind of model and because provider components can be replaced easily, I would be leaning to use this. But I'm also interested to research what are the other ways this can be implemented.
	
> what are the key tasks 
     
- components in core package
- configs and env variables
- Comments either in app layout or specific to Blog layout

> 4. how long will it take guesstimate? (for MVP and polished)
* MVP ~3hrs or more if using some other provider not used by pliny. First pass would be just using the same setup and provider and see if that is working. Then modifying stuff according to our needs eg. based on additional configs.
* polished ~6hrs more or less with design suited to flowershow

khalilcodes:
Closes #163 
### Summary

This PR adds the feature to enable comments on pages by setting them up in config and/or frontmatter.

### Changes
* Create comments components in core for [giscus](https://giscus.app/), [utterances](https://utteranc.es/) and [disqus](https://disqus.com)
  * Add [giscus react](https://github.com/giscus/giscus-component) and [disqus-react](https://github.com/disqus/disqus-react) packages in core 
* Add configs in frontmatter and config.js
  * configured to set page comments through frontmatter, or specified folders in comments' config. If both are not set then defaults to show on all pages. Frontmatter value takes precedence.
* Add documentation @ [docs/page-comments](https://deploy-preview-406--spectacular-dragon-c1015c.netlify.app/docs/page-comments)

### TODO
- [x] add disqus support
- [x] Create documentation
- [x] publish core
- [x] Add giscuss env variables in netlify
- [x] install giscus app in flowershow repo

### Screenshot (giscus comments)
Screen Shot 2023-01-30 at 1 35 11 PM

### DIFF
--- a/.changeset/silver-buses-tap.md
+++ b/.changeset/silver-buses-tap.md
@@ -1,0 +1,5 @@
+---
+"@flowershow/core": patch
+---
+
+Add: Page comments feature with support for three providers - giscus, utterances and disqus.
--- a/.changeset/silver-buses-tap.md
+++ b/.changeset/silver-buses-tap.md
@@ -1,5 +1,5 @@
 ---
-"@flowershow/core": patch
+"@flowershow/core": minor
 ---
 
 Add: Page comments feature with support for three providers - giscus, utterances and disqus.
--- a/.changeset/silver-buses-tap.md
+++ b/.changeset/silver-buses-tap.md
@@ -1,5 +1,0 @@
----
-"@flowershow/core": minor
----
-
-Add: Page comments feature with support for three providers - giscus, utterances and disqus.
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@flowershow/core",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "description": "Core Flowershow components, configs and utils.",
   "repository": {
     "type": "git",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -37,10 +37,5 @@
     "next-themes": "^0.2.1",
     "react": "^18.0.0",
     "react-dom": "^18.0.0"
-  },
-  "overrides": {
-    "kbar": {
-      "react": "^17.0.0"
-    }
   }
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -22,6 +22,7 @@
     "@docsearch/react": "^3.3.0",
     "@floating-ui/react-dom": "^1.2.1",
     "@floating-ui/react-dom-interactions": "^0.13.3",
+    "@giscus/react": "^2.2.6",
     "@headlessui/react": "^1.7.6",
     "clsx": "^1.2.1",
     "framer-motion": "^8.4.3",
@@ -35,5 +36,10 @@
     "next-themes": "^0.2.1",
     "react": "^18.0.0",
     "react-dom": "^18.0.0"
+  },
+  "overrides": {
+    "kbar": {
+      "react": "^17.0.0"
+    }
   }
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -25,6 +25,7 @@
     "@giscus/react": "^2.2.6",
     "@headlessui/react": "^1.7.6",
     "clsx": "^1.2.1",
+    "disqus-react": "^1.1.5",
     "framer-motion": "^8.4.3",
     "kbar": "0.1.0-beta.39",
     "mdx-mermaid": "^1.3.2",
--- a/packages/core/src/ui/Comments/Utterances.tsx
+++ b/packages/core/src/ui/Comments/Utterances.tsx
@@ -6,20 +6,20 @@ export interface UtterancesConfig {
   pages?: Array<string>;
   config: {
     theme?: string;
-    repo?: string;
-    label?: string;
-    issueTerm?: string;
+    repo: string;
+    label: string;
+    issueTerm: string;
   };
 }
 
 export type UtterancesProps = UtterancesConfig["config"];
 
-export const Utterances = ({
+export const Utterances: React.FC<UtterancesProps> = ({
   repo,
   label = "comments",
   issueTerm = "pathname",
   theme = "github-light",
-}: UtterancesProps) => {
+}) => {
   const { theme: nextTheme, resolvedTheme } = useTheme();
   // TODO: remove preferred-color-scheme after theme toggle fix
   const commentsTheme = nextTheme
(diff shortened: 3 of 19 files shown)


packages/core/src/ui/Comments/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Supports Giscus, Utterances
+ * If you want to use a comments provider you have to add it to the
+ * content security policy in the `next.config.js` file.
+ * slug is used in disqus to identify the page
+ *
+ * @param {CommentsProps} { comments, slug }
+ * @return {*}
+ */
```
olayway, line 50: Some old comment?

```suggestion
```

packages/core/src/ui/Comments/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Utterances, UtterancesConfig, UtterancesProps } from "./Utterances";
+import { Disqus, DisqusConfig, DisqusProps } from "./Disqus";
+
+declare global {
+  interface Window {
+    disqus_config?: (...args: any[]) => void;
+    DISQUS?: (...args: any[]) => void;
+    page?: any;
+  }
+}
```
olayway, line 12: Comment out or remove?

packages/core/src/ui/Comments/Utterances.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { useTheme } from "next-themes";
+
+export interface UtterancesConfig {
+  provider: "utterances";
+  pages?: Array<string>;
+  config: {
+    theme?: string;
+    repo?: string;
+    label?: string;
+    issueTerm?: string;
```
olayway, line 11: I'm guessing `theme` and `issueTerm` could be of more strict types, e.g. `theme: "light" | "dark"`?

packages/core/src/ui/Comments/Utterances.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const COMMENTS_ID = "comments-container";
+
+  const LoadComments = useCallback(() => {
+    const script = document.createElement("script");
+    script.src = "https://utteranc.es/client.js";
+    script.setAttribute("repo", repo as string);
+    script.setAttribute("issue-term", issueTerm as string);
+    script.setAttribute("label", label as string);
+    script.setAttribute("theme", commentsTheme as string);
```
olayway, line 39: So are these values always defined? In this case in the interface they shouldn't have `?` signs next to their names.

olayway (changes requested):
Good job @khalilcodes! But there are a few things I'd do differently 😋

packages/core/package.json
```diff
@@ -35,5 +37,10 @@
     "next-themes": "^0.2.1",
     "react": "^18.0.0",
     "react-dom": "^18.0.0"
+  },
+  "overrides": {
+    "kbar": {
+      "react": "^17.0.0"
+    }
```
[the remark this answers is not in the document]
khalilcodes: forgot to remove 😅

khalilcodes:
@olayway for a new feat would this be a minor version bump or keep it as patch ?

olayway (approved):
LGTM :) If everything works and you've tested each provider, then I think we can merge!

olayway:
> @olayway for a new feat would this be a minor version bump or keep it as patch ?

Definitely a minor bump 😃

.changeset/silver-buses-tap.md
```diff
@@ -0,0 +1,5 @@
+---
+"@flowershow/core": patch
```
olayway, line 2: Let's make it a "minor" change

rufuspollock:
@khalilcodes @olayway great work everyone 👏 

2 quick small comments:

- docs page naming: why `page-comments` vs `comments`? `comments` seems simpler and i assume this applies to blog pages as well etc?
- could we get a `created` date on the page
- shall we turn this docs page into a blog post by adding `type: Blog` (seems worth of it!)
- have we turned this on for the site? if so where can i try it out?

PR #426, opened 2023-02-13
67 twitter embed

### LINKED ISSUE #67: Tweet link in markdown shows a preview/embed
See previous effort in web3 https://github.com/life-itself/web3/issues/150

Like we do with youtube ... #88 

## Acceptance

* [x] Pasting a twitter link in its own paragraph at line start results in embed e.g. pasting the following results in an embed/preview of the tweet
* [x] Docs page for this **added at `docs/social-previews`**

```
Blah, blah ...

https://twitter.com/rufuspollock/status/1524012242777395201

```

## Tasks
- [x] research and proposed solution with time estimate **Best option for now is similar implementation to web3 https://github.com/life-itself/web3/issues/150**
- [x] create a tweet embed component
    - with placeholder for loading and deleted/not found tweets **2023/02/13 Done**
- [x] add custom styles for proper tweet embed previews (dark mode) **Only adding this after we fix light/dark switch**
- [x] use component in links for parsing in markdown. **Using customlink**

## Notes

- Twitter embed api is fairly slow and so takes time to load. A placeholder has been added for better ux.
- On further research during development, the embeds (youtube, twitter, etc) load iframes with large htmls which affect the page load and SEO.
- Possible solutions to the above problems:
     -  fetch tweets and render component during build time. This requires getting a twitter api key and recreating custom twitter component that matches the look of the embed. (overwork but blazing fast. see [here](https://static-tweet.vercel.app/)).
     - create remark plugin for twitter embeds based on this [gatsby remark plugin](https://github.com/octahedroid/gatsby-remark-twitter/blob/master/index.js). Although still not sure if this would resolve tweet loading time.

## Analysis
- Fetch tweets and render component during build time not going to work now due to changes in twitter's api
- Embed plugin option not going to be efficient since scripts in the browser are needed
- Best option for now is similar implementation to web3 https://github.com/life-itself/web3/issues/150

PhilippeduPreez:
closes #67 

## Summary
- Twitter URLs in markdown turn into twitter embeds

Preview:

image

## Changes
- `TwitterEmbed` component in core added
- `Customlink` Component updated
- doc added at `docs/social-previews`

## Notes
- Core changes in PR needs to be published for this to work
- Dark/Light theme to be added later when our Theme switch is working

### DIFF
--- a/packages/core/src/ui/Base/TwitterEmbed.tsx
+++ b/packages/core/src/ui/Base/TwitterEmbed.tsx
@@ -34,22 +34,22 @@ export default function TwitterEmbed({ url, ...props }) {
 
   const tweetId = url.split("status/").pop();
 
-  const renderTweet = useCallback(() => {
-    window.twttr.widgets
-      .createTweet(tweetId, ref.current as any, {
-        theme: "dark",
-      })
-      .then((el) => {
-        if (el) {
-          setTweetState(TweetState.LOADED);
-        } else {
-          setTweetState(TweetState.FAILED);
-        }
-      });
-    return window.twttr.widgets.load(ref.current as any);
-  }, [tweetId]);
-
   useEffect(() => {
+    const renderTweet = () => {
+      window.twttr.widgets
+        .createTweet(tweetId, ref.current as any, {
+          theme: "dark",
+        })
+        .then((el) => {
+          if (el) {
+            setTweetState(TweetState.LOADED);
+          } else {
+            setTweetState(TweetState.FAILED);
+          }
+        });
+      return window.twttr.widgets.load(ref.current as any);
+    };
+
     if (!window.twttr) {
       const script = document.createElement("script");
       script.src = twitterWidgetJs;
@@ -59,7 +59,7 @@ export default function TwitterEmbed({ url, ...props }) {
     } else {
       renderTweet();
     }
-  }, [renderTweet]);
+  }, [tweetId]);
 
   return (
     <>
--- a/packages/core/src/ui/Base/TwitterEmbed.tsx
+++ b/packages/core/src/ui/Base/TwitterEmbed.tsx
@@ -1,0 +1,124 @@
+import { useEffect, useCallback, useState, useRef, RefObject } from "react";
+
+const twitterWidgetJs = "https://platform.twitter.com/widgets.js";
+const message = "Loading tweet...";
+
+interface TweetConfig {
+  theme: string;
+}
+
+type TweetState = {
+  isLoading: boolean;
+  message: string | null;
+};
+
+declare global {
+  interface Window {
+    twttr: {
+      widgets: {
+        createTweet: (
+          id: string,
+          ref: RefObject<HTMLDivElement>,
+          options: TweetConfig
+        ) => Promise<any>; // TODO type
+        load: (ref: RefObject<HTMLDivElement>) => void;
+      };
+    };
+  }
+}
+
+export default function TwitterEmbed({ url, ...props }) {
+  const ref = useRef<HTMLDivElement | null>(null);
+  const [state, setState] = useState<TweetState>({
+    isLoading: true,
+    message: message,
+  });
+
+  const tweetId = url.split("status/").pop();
+
+  const renderTweet = useCallback(() => {
+    if (typeof window !== undefined) {
+      window.twttr.widgets
+        .createTweet(tweetId, ref.current as any, {
+          theme: "dark",
+        })
+        .then((el) => {
+          if (el) return setState((prev) => ({ ...prev, isLoading: false }));
+          return setState((prev) => ({
+            ...prev,
+            message: null,
+          }));
+        });
+      return window.twttr.widgets.load(ref.current as any);
+    }
+  }, [tweetId]);
+
+  useEffect(() => {
+    let componentMounted = true;
+
+    const script = document.createElement("script");
+    script.src = twitterWidgetJs;
+    script.async = true;
+    script.onload = () => renderTweet();
+
+    if (componentMounted) {
+      if (!window.twttr) {
+        document.head.appendChild(script);
+      } else {
+        renderTweet();
+      }
+    }
+
+    return () => {
+      setState({ isLoading: true, message: message });
+      componentMounted = false;
+    };
+  }, [renderTweet]);
+
+  return (
+    <>
+      {state.isLoading && state.message && (
+        <div className="relative my-4 w-full sm:max-w-xl bg-neutral-900 drop-shadow-md rounded-lg">
+          <div className="absolute flex flex-col flex-wrap break-all items-center justify-center bg-slate-700/60 w-full h-full px-4 py-2 rounded-lg top-0 left-0 z-10">
+            <svg
+              role="img"
+              viewBox="0 0 24 24"
+              xmlns="http://www.w3.org/2000/svg"
+              className="w-6 absolute right-4 top-4"
+            >
+              <title>Twitter</title>
+              <path
+                fill="#1DA1F2"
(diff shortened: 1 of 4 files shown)


packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+const message = "Loading tweet...";
+
+interface TweetConfig {
+  theme: string;
+}
+
+type TweetState = {
+  isLoading: boolean;
+  message: string | null;
+};
```
olayway, line 13: Can you try using enum for all possible states instead? The message can later be derived based on the current Tweet state. But there is no need to keep it in state. Unless we would e.g. like to display some responses from the server. In our case we're only using "Loading..." or `null`.

packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+    };
+  }
+}
+
+export default function TwitterEmbed({ url, ...props }) {
+  const ref = useRef<HTMLDivElement | null>(null);
+  const [state, setState] = useState<TweetState>({
+    isLoading: true,
+    message: message,
+  });
```
olayway, line 35: Refactor to use enum.

packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const tweetId = url.split("status/").pop();
+
+  const renderTweet = useCallback(() => {
```
olayway, line 39: What's the benefit of using `useCallback` here?

packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+    if (typeof window !== undefined) {
```
olayway, line 40: I don't think you need this check, as you're only running `renderTweet` on the client.

packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+      window.twttr.widgets
+        .createTweet(tweetId, ref.current as any, {
+          theme: "dark",
+        })
+        .then((el) => {
+          if (el) return setState((prev) => ({ ...prev, isLoading: false }));
+          return setState((prev) => ({
+            ...prev,
+            message: null,
+          }));
```
olayway, line 50: This can be simplified. Use enums. And also, you don't need to return, as you're not doing anything with the returned value.

packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+    }
+  }, [tweetId]);
+
+  useEffect(() => {
+    let componentMounted = true;
+
+    const script = document.createElement("script");
+    script.src = twitterWidgetJs;
+    script.async = true;
+    script.onload = () => renderTweet();
```
olayway, line 62: The script should only be created when it's not already there.

packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+        document.head.appendChild(script);
+      } else {
+        renderTweet();
+      }
+    }
+
+    return () => {
+      setState({ isLoading: true, message: message });
+      componentMounted = false;
+    };
```
olayway, line 75: Why do you think it's needed?

packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+  }, [renderTweet]);
+
+  return (
+    <>
+      {state.isLoading && state.message && (
```
olayway, line 80: This doesn't really make sense when I'm reading this. It's not self-explanatory. Why would a message affect displaying a placeholder? Or why the placeholder doesn't need to be displayed if there is no message, even if the state is still loading?

packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+                d="M23.953 4.57a10 10 0 01-2.825.775 4.958 4.958 0 002.163-2.723c-.951.555-2.005.959-3.127 1.184a4.92 4.92 0 00-8.384 4.482C7.69 8.095 4.067 6.13 1.64 3.162a4.822 4.822 0 00-.666 2.475c0 1.71.87 3.213 2.188 4.096a4.904 4.904 0 01-2.228-.616v.06a4.923 4.923 0 003.946 4.827 4.996 4.996 0 01-2.212.085 4.936 4.936 0 004.604 3.417 9.867 9.867 0 01-6.102 2.105c-.39 0-.779-.023-1.17-.067a13.995 13.995 0 007.557 2.209c9.053 0 13.998-7.496 13.998-13.985 0-.21 0-.42-.015-.63A9.935 9.935 0 0024 4.59z"
+              />
+            </svg>
+            <div className="text-gray-300 font-bold my-2 italic">
+              {state.message}
+            </div>
+          </div>
+          <div
+            className={`p-3 space-y-4 ${
+              state.isLoading && state.message === message && "animate-pulse"
```
olayway, line 101: Incorrect use of `&&`.

olayway, line 101: Incorrect use of `&&`.

packages/core/src/ui/Base/TwitterEmbed.tsx
```diff
@@ -0,0 +1,124 @@
...
+            <div className="flex space-x-4">
+              <div className="w-1/4 h-3 bg-slate-700"></div>
+              <div className="w-1/4 h-3 bg-slate-700"></div>
+              <div className="w-1/4 h-3 bg-slate-700"></div>
+            </div>
+          </div>
+        </div>
+      )}
+      <div className="twitter-tweet" ref={ref} />
+      {!state.message && <p {...props} />}
```
olayway, line 121: What is this for?

olayway (changes requested):

olayway:
@PhilippeduPreez additionally, I'd try using Next.js Script component: https://nextjs.org/docs/basic-features/script. Although I'm not sure if it will work here. But take a look at it.

packages/core/src/ui/Base/TwitterEmbed.tsx
PhilippeduPreez: I get the following if not using a useCallback: 

The 'renderTweet' function makes the dependencies of useEffect Hook change on every render. Move it inside the useEffect callback. Alternatively, wrap the definition of 'renderTweet' in its own useCallback() Hook.

olayway: Yes, I know, but the `renderTweet` function doesn't even have to be a dependency of `useEffect`. It only makes the code less readable in this case and is used here like a patch, because sth is showing an error if not done otherwise...
The reason why in this case `renderTweet` has to be listed as a dependency of `useEffect` is that 1) it's called inside the `useEffect` callback AND 2) it's not a pure function, i.e. it has references to props (or state) which we want to trigger the `useEffect`. Instead, we could be passing these props (or state) directly to `useEffect`. 
And if `renderTweet` was a pure* function, i.e. it would be explicitly passed values it needs as arguments (instead of accessing them from the upper scope) yes, on each render another instance of it would be created (which is not a problem in this case) but it would have exactly the same effect when run with the same arguments. And you wouldn't have to pass it then as a dependency of `useEffect.` 

Another way of doing it is what the error suggests to you in the first place: moving this function inside the useEffect callback.

*Note: pure but not in a very strict sense, because it will still have references to `window`, but it the same on each render, so doesn't make a difference.

Let me know if it's still not clear, I didn't want to solve it for you 😛

olayway (changes requested):

olayway (approved):
LGTM :)

PR #440, opened 2023-02-21
Exporting Themeprovider from Core

### LINKED ISSUE #386: Theme switch (dark / light) bug (on our website)
The theme toggle does not switch the theme to dark / light mode as expected.

## Acceptance
- [ ] Theme switch toggle works

## Tasks
- [ ] Find the root cause
- [ ] ...

PhilippeduPreez:
closes #386 

## Summary
- removed `next-themes` package from template
- export `ThemeProvider` from core

## Notes
- Theme switch works locally and build passes
- Want to make sure this does not cause any problems/changes with the CLI tool

### DIFF
--- a/packages/core/src/ui/Base/ThemeProvider.tsx
+++ b/packages/core/src/ui/Base/ThemeProvider.tsx
@@ -1,2 +1,0 @@
-import { ThemeProvider } from "next-themes";
-export { ThemeProvider };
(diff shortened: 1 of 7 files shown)


packages/core/src/ui/Base/ThemeProvider.tsx
```diff
@@ -0,0 +1,2 @@
+import { ThemeProvider } from "next-themes";
+export { ThemeProvider };
```
khalilcodes, line 2: I don't think a separate file is required here, we can directly export ThemeProvider in index.

khalilcodes:
@PhilippeduPreez @olayway we might not need to do this, see comment in issue.

olayway (approved):

PR #442, opened 2023-02-24
Auto extract title and description from content

### LINKED ISSUE #297: Auto-extract title and description for a page
Be nice if we didn't have to specify title and description in frontmatter especially for e.g. blog posts.

Could we auto-extract title as line 1 heading (if it exists) and extract description as first X characters/words or first paragraph of page?

Bonus: also set up "updated" date for a post based on e.g. last modified in git (if running off git!)

Here's an issue showing how to extract the description (aka excerpt): https://github.com/contentlayerdev/contentlayer/issues/76

NB: should we have description and excerpt be different?

## Acceptance

- [ ] if title and/or description already in frontmatter they are used (if non-empty)
- [ ] if no title in frontmatter, and first non-blank line of markdown contains a heading then that heading becomes the title
  - [ ] TODO: may need to strip markdown formatting e.g. if italic, bold, links etc 
- [ ] if no description, the first paragraph of the document (after any title) becomes description (stripped of markdown formatting)

## Tasks

- [x] Resolve empty titles in contentlayer (~1h)  **✅ 2023-02-24**
  - [x] use first h1 heading (if exists)
  - [x] strip markdown formatting
  - [x] handle repeating titles ie. remove first markdown heading if title is displayed (~1h) **✅ 2023-02-26**
- [x] Resolve empty descriptions in contentlayer (~3h) **✅ 2023-02-24**
  - [x] use paragraph/text
  - [x] strip markdown formatting: use https://github.com/remarkjs/strip-markdown
  - [x] strip mdx formatting (commented lines, import statements, ... ?) **using regex**

khalilcodes:
Closes #297 
### Summary

Merging this PR will add meta title and description from content's (first) `H1`  and `paragraph` respectively, only if they are not specified in the page's frontmatter and if page content exists.

### Changes

* Add resolvers for title and description in computed fields - `contentlayer.config.ts`
* Add package - https://github.com/remarkjs/strip-markdown
* Handle repeating titles in `pages/[[...slug]].tsx`

### DIFF
--- a/.changeset/tender-ravens-fetch.md
+++ b/.changeset/tender-ravens-fetch.md
@@ -1,0 +1,5 @@
+---
+"@flowershow/template": minor
+---
+
+Auto extract title and description from content.
--- a/packages/template/contentlayer.config.ts
+++ b/packages/template/contentlayer.config.ts
@@ -1,6 +1,12 @@
 /* eslint import/no-unresolved: off */
-import { defineDocumentType, makeSource } from "contentlayer/source-files";
+import {
+  ComputedFields,
+  defineDocumentType,
+  makeSource,
+  FieldDefs,
+} from "contentlayer/source-files";
 import { h } from "hastscript";
+import { remark } from "remark";
 import rehypeAutolinkHeadings from "rehype-autolink-headings";
 import rehypeMathjax from "rehype-mathjax";
 import rehypePrismPlus from "rehype-prism-plus";
@@ -13,11 +19,11 @@ import remarkToc from "remark-toc";
 import callouts from "@flowershow/remark-callouts";
 import remarkEmbed from "@flowershow/remark-embed";
 import wikiLinkPlugin from "@flowershow/remark-wiki-link";
+import stripMarkdown, { Options } from "strip-markdown";
 
 import { siteConfig } from "./config/siteConfig";
 
-// TODO types
-const sharedFields: any = {
+const sharedFields: FieldDefs = {
   title: { type: "string" },
   description: { type: "string" },
   image: { type: "string" },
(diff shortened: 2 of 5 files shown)


PhilippeduPreez:
@khalilcodes Everything looks great, except since the h1 is made into a title, that means with our current layout setup the heading will display twice, so we will either have to remove the h1 again in a place like `[[slug]]` or check in layout or change how our layout works, which sounds messy. Or if we mainly care about SEO add h1 to an eg. `metatitle` frontmatter, but then It won't display in places like the `Bloglist`

rufuspollock:
> @khalilcodes Everything looks great, except since the h1 is made into a title, that means with our current layout setup the heading will display twice, so we will either have to remove the h1 again in a place like `[[slug]]` or check in layout or change how our layout works, which sounds messy. Or if we mainly care about SEO add h1 to an eg. `metatitle` frontmatter, but then It won't display in places like the `Bloglist`

@PhilippeduPreez good catch. My quick thought @khalilcodes (others may have better ideas). Seem to be me two options:

- Remove the h1 from the markdown. This would be best b/c we may have special styling of the h1 when it comes from title (e.g. we show title, then metadata for blog post etc). My concern is that it is tricky / bug prone. however, if we could do it i would do this. Note: could be a hacky way to do this with css e.g. `display: none` the first h1 after the main title.
- Create a special flag e.g. `flowershow_extracted_title` which we set to true and if that is set we do *not* display title from frontmatter title. This is less error prone possibly. It has disadvantages mentioned in previous point.

olayway (approved):

olayway:
@khalilcodes only one suggestion for the future: I'd create custom remark processors in separate files and use them in contentlayer config file, so that we don't pollute it too much

PR #450, opened 2023-03-02
Custom page and layout tutorial

PhilippeduPreez:
## Summary
- added custom page and layout tutorial


site/content/docs/custom-layout.md
```diff
@@ -0,0 +1,149 @@
...
+Let's start with creating a new `Recipe` document type.
+
+## Create a custom Contentlayer document type
+
+In order to create our new `Recipe` document type, we need to go inside `.flowershow/contentlayer.config.ts` and add instructions to tell Contentlayer how to parse:
+
+- `name`: namespace
+- `filePathPattern`: input files (Files within this path do not need to specify their `Type` to `Recipe`)
+- `fields`: meta data fields
+- `computedFields`: derived meta data fields
```
olayway, line 20: I don't really understand this part 😅 I'd rather expect some intro, saying what metadata and other data will be needed in all of our recipe pages, e.g. prep time, cook time, easy/medium/hard, for how many serves it is, maybe cost, image, and obviously ingredients. Then I'd show how to create a document of this type in contentlayer config.

site/content/docs/custom-layout.md
```diff
@@ -0,0 +1,149 @@
...
+```javascript
+//contentlayer.config.ts
+// ...
+const recipeFields = {
+  created: { type: "date", required: true },
+  authors: {
+    type: "list",
+    of: { type: "string" },
+  },
+  layout: { type: "string", default: "recipe" },
```
olayway, line 31: But we're not suing any new fields here 😭. Take a look at this example recipe page: https://www.bbcgoodfood.com/recipes/classic-lasagne-0. We could have a lot of fancy stuff added to frontmatter that we could then use in the custom layout. We could even have a field for ingredients list so that we can display them in some box on the left or sth like that.

site/content/docs/custom-layout.md
```diff
@@ -0,0 +1,149 @@
...
+---
+title: Homemade Lasagne Recipe
+description: Lasagne is a classic Italian dish that is perfect for any occasion. It is a hearty and delicious meal that can be enjoyed by everyone.
+created: 2022-11-29
+authors: [John Doe]
+---
+Lasagne is a classic Italian dish that is perfect for any occasion. It is a hearty and delicious meal ...
+```
+
+>See more on how to use the `Author` document type [here](https://flowershow.app/docs/blog#blog-authors)
```
olayway, line 64: You can use callouts for extra info like that. Take a look at our docs on callouts/admonitions.

site/content/docs/custom-layout.md
```diff
@@ -0,0 +1,149 @@
...
+
+After creating our `Recipe` document type, we are ready to create a custom layout to display this.
+
+## Create a custom layout
+
+Now, we can create our `RecipeLayout.tsx` at `.flowershow/layouts` to receive and display the title, date and authors for the pages:
+
+```javascript
+export function RecipeLayout({ children, ...frontMatter }) {
+    const { created, title, authors } = frontMatter
```
olayway, line 74: We could extract much more data here... Even all the ingredients so that they can be displayed always in the same place on the recipe page, e.g. some box at the top or on the left hand side. Take a look at some example sites with recipes. I'm not saying you should create something fancy, but just try to show users why creating custom layout may be useful at all 🙂

olayway (changes requested):
Good start, but I don't think this is showing the potential of custom document types and layouts. We're not showing anything more than we already do in the blog tutorial. We're not adding any new metadata on recipe pages and we're not using it in the custom recipe layout.

Also: can you please create an example template for it in our /examples folder and let users know about it somewhere in this tutorial?

PhilippeduPreez:
@olayway Changes made :)

site/content/docs/custom-layout.md
```diff
@@ -0,0 +1,248 @@
...
+---
+title: Great Recipes
+layout: simple
+data:
+  - recipes
+---
+
+<BlogsList blogs={recipes}/>
+```
+
```
olayway, line 243:
```suggestion
- `data: - recipes` is the way we tell Flowershow which getter functions we want to call in order to provide the page with the data it needs. This data will then be available in the whole MDX under the same name, so you'll be able to pass it to your own custom components.

> [!note]
> Note, that the name of the getter listed in the `data` frontmatter field (in our case `recipes`) needs to be exactly the same as the name of the getter function used to fetch your custom-type documents.
```

site/content/docs/custom-layout.md
```diff
@@ -0,0 +1,248 @@
...
+An additional benefit of this approach, is that if in the future we would like to change the way our recipes look, we can do this just by changing the default layout for this document type, instead of doing this for every single recipe post.
+
+The last step we have is to create a home page where all our recipes will be displayed.
+
+## Create a home page
+
+In order to list all our recipes on a given page, we can use our `BlogsList` component and pass it a list of our recipes. It's globally available, so we don't need to import it.
+
+First we need to create a getter function that will fetch all our recipes:
```
olayway, line 221:
```suggestion
But first, we need to create a getter function that will fetch all our recipes:
```

olayway (changes requested):
Much better but I've suggested some improvements.

olayway (approved):

PhilippeduPreez:
@olayway Latest updates made!

PR #453, opened 2023-03-06
Make date the default field in frontmatter

### LINKED ISSUE #445: Support for `date` frontmatter field as default date field rather than `created`
`date` frontmatter field is default date field rather than `created`

See discussion #416

## Acceptance

- [x] `date` frontmatter field support added
- [x] `created` frontmatter filed supported as well for backwards compatibility

## Tasks
- [x] add `date` computed field in contentlayer's config file
  - [x] takes frontmatter's `date` field value if present
  - [x] otherwise uses frontmatter's `created` field value

PhilippeduPreez:
closes #445 

## Summary
- Uses `date` as default frontmatter field and uses `created` field if `date` is not found

## Changes
- Blog layout components updated to use `date` instead of `created`
- `date` computed field added in `contentlayer.config`

## Notes
- Since either `created` or `date` can be used, none of them is required, maybe there is a way to have at least one of them required?
- Tested that both date and created frontmatter fields works as expected

### DIFF
--- a/.changeset/khaki-toys-crash.md
+++ b/.changeset/khaki-toys-crash.md
@@ -1,0 +1,6 @@
+---
+"@flowershow/template": patch
+"@flowershow/core": patch
+---
+
+Make date frontmatter the default field being used, created can still be used in frontmatter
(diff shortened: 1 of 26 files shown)


rufuspollock:
@khalilcodes could you review and approve if it looks good.

rufuspollock:
@olayway let you review as well now you are around.

PR #456, opened 2023-03-07
Link previews per page added

### LINKED ISSUE #455: Enable/disable hover-over tooltips per page
I want the option to enable/disable hover-over tooltips per page, so I can have more creative freedom on how my pages are displayed

## Acceptance
- [x] hover-over tooltips can be enabled/disabled per page **added showLinkPreview frontmatter field**
- [x] remove hover-over tooltip from flowershow's home page and index pages **Done**
- [x] add change to docs **https://flowershow.app/docs/link-previews**

## Tasks
- [x] hover-over tooltips can be enabled/disabled per page
- [x] remove hover-over tooltip from flowershow's home page and index pages
- [x] add change to docs

PhilippeduPreez:
closes #455 

Link previews boolean added to frontmatter

## Changes
- frontmatter value `showLinkPreview` will be used first if it exists instead of siteConfig value `showLinkPreviews`
- added change to docs at  `/docs/link-previews`
- changed our home page and docs index page to not display link previews

### DIFF
--- a/.changeset/new-doors-fetch.md
+++ b/.changeset/new-doors-fetch.md
@@ -1,0 +1,5 @@
+---
+"@flowershow/template": patch
+---
+
+Link previews can be enabled/disabled per page
--- a/packages/template/contentlayer.config.ts
+++ b/packages/template/contentlayer.config.ts
@@ -33,6 +33,7 @@ const sharedFields: FieldDefs = {
   showSidebar: { type: "boolean" },
   showComments: { type: "boolean" },
   isDraft: { type: "boolean" },
+  showLinkPreview: { type: "boolean" },
   data: { type: "list", of: { type: "string" }, default: [] },
 };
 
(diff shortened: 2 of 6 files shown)


rufuspollock:
@PhilippeduPreez why are we implementing this feature? This will add complexity to the app and it seems very specific need - what was the use case for this feature?

khalilcodes (approved):
@PhilippeduPreez check comment from rufus ☝️ before merge and if we really need this.

olayway:
> @PhilippeduPreez why are we implementing this feature? This will add complexity to the app and it seems very specific need - what was the use case for this feature?

@rufuspollock The use case is our home page, where we originally had links to features' doc pages under "Read more" and I don't think it looked good. Now we only have one wiki link on that page, so this could probably wait but tbh I'm glad we have this option now, as I personally wouldn't like to see tooltip previews on my personal site's home page or on any index pages.

Not sure what is the complexity added here. We have options to enable/disable things per-page for anything else and it's one line of code. And it probably took 5 min of work.

PR #467, opened 2023-03-28
Remark wiki link refactor

### LINKED ISSUE #464: `remark-wiki-link` package refactoring
## Acceptance
- [x] supports all of the formats below
  - [x] [[Wiki Link]]   # relative wiki link
  - [x] [[Wiki Link|Alias]]   # wiki link with alias
  - [x] [[some/folder/Wiki Link]] # obsidian absolute paths
  - [x] [[Wiki Link]] # obsidian shortened paths
  - [x] ![[...]]  # for images and other embeds **🚧 some small bug in the syntax extension, but I'm too tired to spot it 😅**
- [x] has extensive tests for
  - [x] `syntax` extension
  - [x] `html` extension
  - ~[ ] `fromMardown` extension~ **removed, not needed**  
  - ~[ ] `toMarkdown` extension~ **removed, not needed**  
  - [x] `remark-wiki-link` plugin itself
- [x] replace `mocha` with `jest` as it's used in all other projects
- [x] refactor to TS

## Tasks
- [x] understand how the plugin and all the micromark extensions work 
- [ ] ...

olayway:
Closes #464

## Changes
- `@flowershow/remark-wiki-link`
  - general code refactoring: tidying up, simplifying, using better naming etc.
  - disable passing markdown folder in plugin config -> export `getPermalinks` function, so that users (e.g. our template) can run it on their side and then use a standard config option `permalinks` to explicitly pass the permalinks list
  - extend tests coverage
  - refactor to TS
  - replace `mocha` with `jest`
- `@flowershow/template`
  - import `getPermalinks` from `@flowershow/remark-wiki-link`, generate the permalinks list from the content folder on the template side (in `contentlayer.config.ts`), and pass it explicitly to the plugin instead of passing the path to markdown folder to it

## Notes
I think there should be a way of using `micromark` extensions only - our `html` and `syntax` -  but maybe I'm missing sth (still not quite getting all the different elements of the system and ecosystem...) and we actually need to use `syntax` extension for `micromark` combined with `fromMarkdown` and `toMarkdown` for `mdast-util-from-markdown`. I actually started by writing `syntax` and `html` and both of them used in `micromark` do their job, i.e. they correctly convert md to html (see tests in `remark-wiki-link/tests/micromarkExtensionWikiLink.spec.ts` - mirror tests of `fromMarkdown`). But I couldn't make it work without using `fromMarkdown` - **which is almost identical to `html`** - and `toMarkdown`.
For reference see also `remark-wiki-link/src/lib/remarkWikiLink.ts` file.


rufuspollock (approved):
Generally looks good (have high confidence in your code, there's a lot here and not sure i get all of it so have not reviewed in detail).

My one big comment would be there is no change to the README.md - have we changed the API (it looks like we have looking at `contentlayer.config.js`)? If so, worth noting that.

olayway:
> Generally looks good (have high confidence in your code, there's a lot here and not sure i get all of it so have not reviewed in detail).
> 
> My one big comment would be there is no change to the README.md - have we changed the API (it looks like we have looking at `contentlayer.config.js`)? If so, worth noting that.

Oh, yes! You're right! I'll add notes to readme!

rufuspollock:
@olayway i think this is good to merge (or maybe you are waiting on README)

one small thing: when you merge could you copy and paste the excellent summary in PR into the body of the merge commit message?

olayway:
> @olayway i think this is good to merge (or maybe you are waiting on README)

Yes, I'm just finishing adding configuration options to README and I'm merging

> one small thing: when you merge could you copy and paste the excellent summary in PR into the body of the merge commit message?

Will do! :)

PR #475, opened 2023-04-05
MarkdownDB fixes

olayway:
## Changes

**`package.json`**
- added missing dependencies
- added `mddb` command that executes folder indexing (see `/src/bin/index.js`) - will probably be replaced with a full-blown CLI in the future

**`/lib`**
- introduced the distinction between simple object interfaces (e.g. `File`) from instances of database rows (e.g. `MddbFile`) represented by `Mddb...` classes that implement additional methods and in general should only be instantiated from records returned from the database
- added a simple validation function that checks if some of the objects in an array have the same value of a given property
  - added this check in `batchInsert` methods on `Mddb...` classes to catch duplicate values in JS and display objects that would cause SQL errors (which are impossible to debug 😅)
  - maybe we could add some validation package in the future, but for now this should do

\+ Bunch of other fixes caught after testing with `datahub-next`


demenech (approved):
LGTM

PR #476, opened 2023-04-06
Markdowndb README v.1

### LINKED ISSUE #471: MarkdownDB README with features, motivation, brief tutorial
New user: When coming to MarkdownDB I want to know (briefly) what it is, what features it has (existing or future) and a brief tutorial on use.

We could port material from https://datahub.io/notes/markdowndb especially the job stories (they could become features section).

- Note one clarification re https://datahub.io/notes/markdowndb is that there is an ambiguity in use between markdowndb as the database itself and markdowndb as a library (index and API) for accessing a markdowndb. Our package is the latter and so for us the description of "markdowndb" is something like: "A javascript library for treating a directory of markdown files as a database" (or a better version thereof).

Emphasize: fine if this is a rough cut, we will rapidly iterate. better to get something down ...

Also find starting from a tutorial is something easier (a short example is worth a thousand descriptions) eg.

````md
On disk you have:

```
my-blog-folder-of-markdown-files/
  blog-1.md
  blog-2.md
```

Then do ...

```
import db from markdowndb

db.create('path-to-folder')
db.query('all blogs written by x author'_
```

Behind the scenes we are creating an SQL(ite) database so you can do everything sql can! Here are some examples ...

etc
````

## Acceptance

- [x] Brief description of what markdowndb is and why you might use it (e.g. why we will use it in Flowershow etc) **see https://github.com/flowershow/flowershow/tree/main/packages/markdowndb#markdowndb**
- [x] Brief description (e.g. bullet list) of features existing or forthcoming (emojis are a bonus!) **see https://github.com/flowershow/flowershow/tree/main/packages/markdowndb#features**
- [x] Quick start tutorial **see https://github.com/flowershow/flowershow/tree/main/packages/markdowndb#quick-start**

### PR DESCRIPTION
**Pull request by olayway**
Closes #471

### DIFF
--- a/packages/markdowndb/README.md
+++ b/packages/markdowndb/README.md
@@ -1,3 +1,109 @@
-# markdowndb
+# MarkdownDB [![](https://badgen.net/npm/v/@flowershow/markdowndb)](https://www.npmjs.com/package/@flowershow/markdowndb)
 
-Parse markdown files and store them in an SQL database.
+MarkdownDB is a ... that parses your markdown files and saves them to your local SQLite database, so that you can easily import them into your application.
+
+**🚧 MarkdownDB is in its early development stage**
+
+## How to use in Next.js app
+
+If you don't have your Next.js project yet, create one with:
+
+```sh
+npx create-next-app@latest my-mddb-app
+```
+
+Once the project is created, install the MarkdownDB package:
+
+```sh
+npm i @flowershow/markdowndb
+```
+
+And run the following command, passing the path to your markdown content folder as an argument.
+
+```sh
+npx mddb <path-to-your-content-folder>
+# e.g.
+# npx mddb my-mddb-content
+```
+
+🎉 Done! You should see a new `markdown.db` file created in the root of your project.
+
+You can preview it with any SQLite viewer, e.g. https://sqlitebrowser.org/.
+
+You can also add it as a script to your `package.json` file:
+
+```json
+{
+  "name": "my-mddb-app",
+  "scripts": {
+    ...
+    "mddb": "mddb <path-to-your-content-folder>",
+    "prebuild": "npm run mddb"
+  },
+  ...
+}
+
+```
+
+Now, once the data is in the database, you can add the following script to your project (e.g. in `/lib` folder). It will allow you to establish a single connection to the database and use it across you app.
+
+```
+// lib/mddb.ts
+import { MarkdownDB } from "@flowershow/markdowndb";
+
+const dbPath = "markdown.db";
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/markdowndb/README.md [hunk 1]
```diff
@@ -1,3 +1,109 @@
-# markdowndb
+# MarkdownDB [![](https://badgen.net/npm/v/@flowershow/markdowndb)](https://www.npmjs.com/package/@flowershow/markdowndb)
 
-Parse markdown files and store them in an SQL database.
+MarkdownDB is a ... that parses your markdown files and saves them to your local SQLite database, so that you can easily import them into your application.
```
rufuspollock, line 3: Suggest something like:

MarkdownDB is a javascript library for working with a "MarkdownDB" i.e. a simple way to treat a directory of markdown files as a database.

Note: this is obviously moving the work into defining a MarkdownDB. Having tried to explain the idea to a few people over the last week or so i think this is the right way to go i.e. define a MarkdownDB and then this *library* is simply a node/js library for working with a markdowndb (it's also the reference API for now etc).

So what's a MarkdownDB? See WIP at https://github.com/flowershow/flowershow/discussions/477

packages/markdowndb/README.md
```diff
@@ -1,3 +1,109 @@
...
+
+**🚧 MarkdownDB is in its early development stage**
+
+## How to use in Next.js app
```
rufuspollock, line 7: Good start. Personally i think i'd leave use in nextjs app second and have first a really simple intro like:

- create the index
- boot up sqlite and do some queries, find me all docs with these tags (i don't know if we have a demo db we can use)

Then tutorial 2 is: building a blog in nextjs (or even just: here's how you'd use it to get stuff in `getStaticProps`)

packages/markdowndb/README.md
```diff
@@ -1,3 +1,109 @@
...
+        filename: dbPath,
+    },
+});
+
+const clientPromise = client.init();
+
+export default clientPromise;
+```
+
+Now, in any of your pages you can use it to query the database, e.g.:
```
rufuspollock, line 68: Cool, this is really great.

rufuspollock (approved):
LGTM - i've left a few comments but we could merge as is and rapidly iterate.

PS: i'm increasingly thinking we want to split this out as a separate lib from Flowershow but that can wait ...

packages/markdowndb/README.md
rufuspollock: NB: I see i already had an example like this in the body of the issue at #471

PR #564, opened 2023-09-19
🛠️fix: footer position for page with short content

siam1113:
Resolves https://github.com/datopian/flowershow/issues/563


rufuspollock:
LGTM

PR #620, opened 2024-01-23
[][m]: Implement hot-reloading

mohamedsalem401:
For https://github.com/datopian/flowershow/issues/587
- It watches for file changes in the `content` folder.
- Using `next-remote-watch`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@types/react": "18.0.28",
         "@types/react-dom": "18.0.11",
         "clsx": "^2.1.0",
+        "concurrently": "^8.2.2",
         "date-fns": "^2.29.3",
         "gray-matter": "^4.0.3",
         "hastscript": "^7.2.0",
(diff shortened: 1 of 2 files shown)


olayway (changes requested):
This is a great start and I would definitely use this package as well. But we also need a way to re-index changed files (to get current metadata) and index new ones (so that they can be rendered at all). Fortunately, this package allows for running custom scripts on change/add/delete/... events with `--script` flag! And the cool thing is that the script will be passed the event type as an argument, so in the future we could use it to only index/re-index/delete single files. For now, let's just re-index the whole folder on any type of the event. Then you can try implementing this the right way.

mohamedsalem401:
> This is a great start and I would definitely use this package as well. But we also need a way to re-index changed files (to get current metadata) and index new ones (so that they can be rendered at all). Fortunately, this package allows for running custom scripts on change/add/delete/... events with `--script` flag! And the cool thing is that the script will be passed the event type as an argument, so in the future we could use it to only index/re-index/delete single files. For now, let's just re-index the whole folder on any type of the event. Then you can try implementing it the right way.

An issue with the library is preventing me from using the --command flag
```
error - unhandledRejection: TypeError [ERR_INVALID_ARG_TYPE]: The "file" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:399:5)
    at validateString (node:internal/validators:163:11)
    at normalizeSpawnArguments (node:child_process:545:3)
    at spawn (node:child_process:750:13)
    at FSWatcher.<anonymous> (C:\Datopian\flowershow\node_modules\next-remote-watch\bin\next-remote-watch:64:13)
    at FSWatcher.emit (node:events:512:28)
    at FSWatcher.emitWithAll (C:\Datopian\flowershow\node_modules\chokidar\index.js:540:8)
    at FSWatcher._emit (C:\Datopian\flowershow\node_modules\chokidar\index.js:632:8)
    at listener (C:\Datopian\flowershow\node_modules\chokidar\lib\nodefs-handler.js:370:20) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

olayway:
@mohamedsalem401 Just please make sure the E2E tests are passing and we can merge.

PR #626, opened 2024-01-31
Add user config type

### LINKED ISSUE #625: User config has no type
This makes it quite difficult to know what's supported when hosting Flowershow yourself.

homostellaris:
Adds a type for the user config which makes it easier to know what's possible when setting up your own Flowershow instance.

Closes #625

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules/
 out/
 .obsidian/
 .markdowndb/
+tsconfig.tsbuildinfo
--- a/config/siteConfig.ts
+++ b/config/siteConfig.ts
@@ -1,8 +1,14 @@
-import { defaultConfig } from "@portaljs/core";
 import userConfig from "@/content/config.mjs";
+import {
+  AnalyticsConfig,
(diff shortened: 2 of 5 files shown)


mohamedsalem401:
@homostellaris 
Thanks a lot for this.🙌 I've reviewed and tested this. 

I think it's good to be merged.

mohamedsalem401:
@homostellaris 
The only thing left would be to move this to the right repo, which is https://github.com/datopian/portaljs/blob/main/packages/core/src/config/default.ts, as I don't think those changes should be in Flowershow.

homostellaris:
@mohamedsalem401 I don't think that would make sense as it's describing the flowers how config which is different to the portal config. You'll notice in siteConfig.ts that where appropriate the portal config types are being referenced but you can also see the additional properties that are specific to flowershow.

If you're worried about the failing build I think that's just due to some upstream changes that I need to include in the config.

homostellaris:
@mohamedsalem401 the above commit should fix the build error in Vercel.

Issue #636, opened 2024-03-06
YAML properties with hyphens cause Build error


joesmusings:
Discussed in https://github.com/datopian/flowershow/discussions/632

## Issue
YAML properties that contain a hyphen cause a build error

Example properties:

```
dg-publish: true
dg-publish: true
kanban-plugin: basic
```

## Example Error

```
SyntaxError: Arg string terminates parameters early
    at new Function (<anonymous>)
    at file:///vercel/path0/node_modules/next-mdx-remote/dist/index.js:64:35
    at Object.Kc [as useMemo] (/vercel/path0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:60:240)
    at exports.useMemo (/vercel/path0/node_modules/react/cjs/react.production.min.js:25:208)
    at MDXRemote (file:///vercel/path0/node_modules/next-mdx-remote/dist/index.js:52:21)
    at Wc (/vercel/path0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:68:44)
    at Zc (/vercel/path0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:70:253)
    at Z (/vercel/path0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:76:89)
    at $c (/vercel/path0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:78:98)
    at Zc (/vercel/path0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:71:145)
Error occurred prerendering page "/Home". Read more: https://nextjs.org/docs/messages/prerender-error
SyntaxError: Arg string terminates parameters early
    at new Function (<anonymous>)
    at file:///vercel/path0/node_modules/next-mdx-remote/dist/index.js:64:35
    at Object.Kc [as useMemo] (/vercel/path0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:60:240)
    at exports.useMemo (/vercel/path0/node_modules/react/cjs/react.production.min.js:25:208)
    at MDXRemote (file:///vercel/path0/node_modules/next-mdx-remote/dist/index.js:52:21)
    at Wc (/vercel/path0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:68:44)
    at Zc (/vercel/path0/node_modules/react-dom/cjs/react-dom-server.browser.production.min.js:70:253)
[...]

PR #643, opened 2024-04-28
Fix YAML properties with hyphens cause Build error

mohamedsalem401:
For issue #636:

- After debugging, I discovered that `MDX-Remote` is the cause of this failure.
- It seems to use YAML properties within a `new function(arg1, arg2)` format, where each argument corresponds to a YAML property. This requires the properties to be in a format that's a valid variable name (i.e., no spaces or strange characters).
- `MDX-Remote` replaces variables in the content body with these YAML properties, like so:
```
---
my_name: "Mohamed"
---
My name is {my_name} // the output will be My name is Mohamed
```
- This pull request replaces all hyphens/other characters in the YAML properties with underscores.
- However, it doesn't handle cases where a user employs the YAML property as a variable value. Since this PR doesn't modify the content of the file itself (i.e., it won't change hyphens in the body `{my-name}` to underscores `{my_name}`), this scenario remains unaddressed.

### DIFF
--- a/lib/markdown.ts
+++ b/lib/markdown.ts
@@ -18,6 +18,19 @@ import { serialize } from "next-mdx-remote/serialize";
 
 import siteConfig from "@/config/siteConfig";
 
+function sanitizeKeys(obj: Record<string, any>): Record<string, any> {
+  const sanitizedObj: Record<string, any> = {};
+
+  for (const key in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, key)) {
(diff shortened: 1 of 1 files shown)


rufuspollock:
I'm not clear if this actually changes the file before processing by our mdx pipeline which will cause a bunch of issues - where does the modification happen?

rufuspollock:
WONTFIX as never got clear answer on whether caused issues plus not that important to do. Please re-open if you think differently 🙂

Issue #678, opened 2025-02-16
Optimize images using e.g. nextjs image support plus cloudflare


rufuspollock:
This should be for Cloud but useful to know for self-hosted.

As we are a NextJS app with cloudflare as our default backend that is the way we should look.

Why: we want images to be optimized - at least the local ones (we obviously can't do remote ones)

## Tasks

- [ ] Do some research **🚧2025-02-16 **

## Research

https://www.reddit.com/r/nextjs/comments/1e9eg1s/optimize_image_when_running_nextjs_on_cloudflare/

The complexity for us is that the images are going to be on the r2 storage domain ...

rufuspollock:
@olayway can we get into pasting (informal) shaping in issue description for these items (or do you just shape/implement at same time atm?)

BTW i was thinking we could use cloudflare pretty easily e.g. see this proposed architecture https://developers.cloudflare.com/reference-architecture/diagrams/content-delivery/optimizing-image-delivery-with-cloudflare-image-resizing-and-r2/

My one concern atm is cost - especially whether we may need to limit how many images people have on their free plan (and/or how much traffic they get) to make this sustainable.

olayway:
@rufuspollock Please take a look and let me know what you think. 

https://github.com/flowershow/flowershow-app/blob/main/docs/rfcs/image-optimization.md

oprimon:
Existing flower pages "broke" by displaying images full size instead of the configured size in md syntax (use case: sync obsidian files with Flowershow).

![[image.jpg|200]] gets displayed in fullsize.

marcusyoung:
The same issue occurs when using: `![200](image.jpg)`.

rufuspollock:
@oprimon any chance you can get a reproduction as a pull request on https://github.com/flowershow/demo in the sandbox folder? we'll try to get this fixed asap

olayway:
@oprimon & @marcusyoung thanks for reporting. The issue has just been fixed.

olayway:
This has been largely implemented. But one outstanding thing that would make it even better is storing image geometry at upload time to enable deterministic image rendering - #978.

olayway:
FIXED

PR #984, opened 2025-12-17
RFC: image optimization

olayway:
Refs #678

### DIFF
--- a/rfcs/image-optimization.md
+++ b/rfcs/image-optimization.md
@@ -1,0 +1,276 @@
+---
+date: 2025-12-16
+---
+
+# Image Optimization
+
+## TL;DR
+
+* **We already partially use Vercel image optimization** via `next/image` in layout components (hero, header, featured images, logos).
+* **The biggest current problem** is that **Markdown-rendered content images still use plain `<img>`**, so they:
+  * have no `srcset` / `sizes`
+  * always download the original image
+  * hurt bandwidth, LCP, and overall performance
+* **The most important first step (v1)** is to:
+  * render **all content images with `next/image`**
+  * consistently define correct `sizes` across the app
+* **v2 (optional, later)**: move image transformations & caching from Vercel to Cloudflare (custom loader + R2), if/when we want more control or lower cost.
+
+**Recommendation:**
+👉 **Do v1 first**. It delivers the majority of performance wins with minimal complexity and builds on what we already have.
+
+>[!note]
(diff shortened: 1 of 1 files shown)


rfcs/image-optimization.md
```diff
@@ -0,0 +1,276 @@
...
+---
+
+## Situation
+
+Flowershow serves all user assets (including images) from Cloudflare R2.
+However, **images are currently rendered via two different pipelines**:
+
+1. **Layout-level images** (hero, header, featured image, logo, etc.)
+   * Rendered with `next/image`
+   * Already benefit from **Vercel’s built-in image optimization**
```
rufuspollock: how do they benefit from that if on r2? I'm not getting how this is optimized without knowing image size beforehand.

can you say a bit more.

rfcs/image-optimization.md
```diff
@@ -0,0 +1,276 @@
+---
+date: 2025-12-16
+---
+
+# Image Optimization
+
+## TL;DR
+
+* **We already partially use Vercel image optimization** via `next/image` in layout components (hero, header, featured images, logos).
```
rufuspollock: i guess what i don't get is do we get *any* optimization from vercel image if we do *not* know image actual sizes (which i thought we don't - and i thought we'd need cloudflare images or whatever to have b/c it would do the resizing).

If we do get a benefit then: yes, let's do just option 1 for now as super fast and easy. then we can look at part 2 later.

rufuspollock (commented):
as i comment inline: i guess what i don't get is do we get *any* optimization from vercel image if we do *not* know image actual sizes (which i thought we don't - and i thought we'd need cloudflare images or whatever to have b/c it would do the resizing).

If we do get a benefit then: yes, let's do just option 1 for now as super fast and easy. then we can look at part 2 later.

rufuspollock:
@olayway did you see my comments? If you can respond / resolve we can then merge ...

olayway:
> @olayway did you see my comments? If you can respond / resolve we can then merge ...

Replied. Let me know if it's clearer know so I'll integrate my comments into the rfc and we merge.

rfcs/image-optimization.md
rufuspollock: this is a great clarification and should go into the rfc. we are solving 2 issues here:

- preventing layout shifts
- reducing size of images served to users (good for their experiences and web vitals).

Let's distinguish this a bit in the rfc.

PR #1125, opened 2026-03-02
fix: respect explicit target attribute on external HTML links

### LINKED ISSUE #1123: Respect explicit target on HTML links
Flowershow uses a rehype plugin that makes external links open in a new tab (target="_blank"). This is intentional so readers don’t lose their place.

However, the plugin currently also overrides raw HTML links, even when a user explicitly sets a target (e.g. `target="_self"`).

## Expected behavior

If a link already has a target attribute, Flowershow should respect it.

Example:

`<a href="https://example.com" target="_self">Link</a>` → should open in the same tab

## Proposed fix

Only set `target="_blank"` if the link does not already define target.

nthbotast:
## Summary
- preserve author-defined `target` on external raw HTML links in `rehypeHtmlEnhancements`
- continue adding `target="_blank"` only when no target is provided
- add unit tests covering default behavior and explicit target preservation

## Testing
- `corepack pnpm exec vitest run --project=unit lib/__tests__/rehype-html-enhancements.test.ts`

Fixes #1123

### DIFF
--- a/apps/flowershow/lib/__tests__/rehype-html-enhancements.test.ts
+++ b/apps/flowershow/lib/__tests__/rehype-html-enhancements.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, it } from 'vitest';
+import { unified } from 'unified';
+import remarkParse from 'remark-parse';
+import remarkRehype from 'remark-rehype';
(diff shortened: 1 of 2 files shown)


olayway:
@nthbotast thanks for contributing!

Issue #1038, opened 2026-02-02
Document available CSS variables (and how to use them to customize your theme)


olayway:

rufuspollock:
FIXED. see https://flowershow.app/docs/custom-styles

rufuspollock:
@Abeelha another one for a short changelog or including in a larger changelog for the week.

Abeelha:
@rufuspollock got it, if we have docs/blog on this for more context, would be appreciated! if not, no problem, i can find my way around it and make it

PR #1137, opened 2026-03-04
docs: document available theme CSS variables

nthbotast:
## Summary\n- expand the Custom styles docs with the main theme CSS variables users can override\n- document color, typography, and semantic shade tokens exposed by the default theme\n- add practical  examples for common customization scenarios\n\n## Why\nCloses #1038\n\nThis gives users a concrete list of variables and a safe override strategy without digging through source CSS files.


rufuspollock:
@nthbotast thanks for the contribution and a useful one 👏

PR #1142, opened 2026-03-04
Removed BETA from password protection form

AbdullahAhmadAAK:
The given task was resolved.

"BETA" has been removed from password protection form which was reported to be in site-password-form.tsx.


olayway:
Already done

PR #1215, opened 2026-03-24
content: migrate two legacy Flowershow tutorials from portaljs.com

### LINKED ISSUE #825: Move all flowershow stuff from portaljs and add redirects
When just searching for stuff for flowershow i ended up on

https://www.portaljs.com/blog/edit-a-website-locally

Even if out of date we should migrate this over and add redirects from portaljs.com - and we can add deprecation notices at the top if necessary.

Abeelha:
Closes #825

## Summary

- Copies `create-a-website-from-scratch.md` and `edit-a-website-locally.md` from portaljs.com to `content/flowershow-app/blog/`
- Adds a `> [!warning]` deprecation callout at the top of each post, explaining these were written for the self-hosted Flowershow template (Vercel + GitHub) and linking readers to `/docs/getting-started`
- Updates frontmatter: `created` → `date`, removes `filetype: blog`, adds `description`

## Notes

- Images referenced via `![[/static/img/...]]` were hosted in the portaljs repo — they will show as broken in the migrated posts. The deprecation notice makes clear the content is legacy, so this is acceptable.
- The companion PR to remove the source files and add redirects from portaljs.com is at https://github.com/datopian/portaljs/pull/1493


rufuspollock (changes requested):
@Abeelha i don't see any of the images migrated over here. e.g. /static/image/blog/xxx paths.

Also i think you probably want the tango images locally (they may disappear at some point).

Abeelha:
@rufuspollock downloaded all images and placed them locally, updated.
  - 61 new image files (31 local + 30 Tango downloads), all resized to max 900px wide
  - Both blog posts updated to use local ![[filename]] wikilinks
  - No existing assets touched

PR #1219, opened 2026-03-26
blog: update custom CSS guide to use CSS variables

### LINKED ISSUE #853: Update the blog post about theming (after last HTML structure and styles adjustments)
Inline Tailwind classes has been replaced with single BEM-like semantic classnames (one per HTML element), and the HTML structure itself has been changed slightly too. Our blog post on customizing styles should be updated to explain the new approach to styling your Flowershow site (and mention that you can use official themes collection, and submit your theme there too for others to use).

Abeelha:
## Summary

- Replaces outdated `.bg-background` Tailwind class selector in the background color example with the current CSS variables approach (`--color-l-background` in `:root`)
- Adds a tip callout pointing users to the Themes docs for quick one-line styling
- Adds a note linking to [[custom-styles]] for the full list of CSS variables

Closes #853

### DIFF
--- a/content/flowershow-app/blog/how-to-customize-style.md
+++ b/content/flowershow-app/blog/how-to-customize-style.md
@@ -6,6 +6,9 @@ authors:
   - olayway
 ---
 
+> [!tip] Looking for a quick win?
+> Before writing custom CSS, check out [[themes|Themes]] — Flowershow offers ready-made themes (Letterpress, Superstack, LessFlowery, Leaf) you can apply in one line of config.
+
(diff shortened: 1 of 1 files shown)


olayway (commented):
This is not the only variable we have. We need a more comprehensive documentation. I think there even is an open PR for that.

Abeelha:
@olayway  the PR links to `"[[custom-styles]]"` for the full variable reference at the end of the example. the change is just fixing the outdated` .bg-background` selector **(which no longer works)** not trying to be the comprehensive docs. If there is an open PR for comprehensive variable docs, happy to close this one if it overlaps.

PR #1357, opened 2026-08-13
fix(flowershow): sanitize config on read + validate on write to prevent 500s

olayway:
## Problem

`f74b9ff` guarded one field (`footer.navigation`) after a malformed value 500'd every page of a site. But the site config has ~30 fields, and **any** malformed field in `config.json` (file) or `site.configJson` (DB) can crash the public site the same way — the config is read in the layout, so one bad field = HTTP 500 on every page. Config was validated on neither read nor write (beyond that one field).

This extends the `f74b9ff` pattern to the whole config, on both the read and write side.

## Approach

**One strict Zod schema** models every `SiteConfig` field (incl. the polymorphic/deprecated landmines: `umami` string|object, `theme` string|object, `hero` boolean|object, the `nav.links` `NavLink|NavDropdown` union). It drives two derived behaviors:

- **Write (`validateConfigPatch`)** — strict + partial. Rejects bad shapes **and** unknown keys (`BAD_REQUEST`) on every dashboard save path, not just the editor. Safe to be strict because the only writer of `configJson` is the machine-generated dashboard UI. Clearing a field with `null`/`undefined` and partial nested patches (`{nav:{title:null}}`, `{umami:{src:null}}`) always pass.
- **Read (`sanitizeSiteConfig`)** — a lenient sanitizer that **never throws**. Salvages valid elements from list fields (`footer.navigation`, `nav.links`, `nav.social`, `social`, `redirects`), drops malformed scalar/object fields wholesale, passes unknown/future keys through, and `console.warn`s each drop (with `siteId` + field path) — silent to the visitor.

**`resolveSiteConfig` is now the single pipeline:** `sanitize(db) → sanitize(file) → deepMerge (file wins) → resolve links (opt-in via siteHostname)`. Each source is sanitized *before* merge, so garbage in one source can't wipe a valid field in the other. The inline link-rewriter moved out of `getConfig` into `resolveSiteConfig`.

## Intended behavior changes (both wanted)

- **DB-sourced links now get resolved too** (the rewriter runs post-merge instead of on file-config only). Idempotent for already-absolute URLs, so stored `publicUrl`s are untouched.
- A malformed `config.json` now **falls back to DB config** instead of nulling the whole config.

No React error boundary — read-side safety relies on the sanitizer + passthrough. Existing footer runtime guards kept as belt-and-suspenders.

## Tests

Written failing-first (per `f74b9ff`). New coverage: the strict validator (bad shapes, unknown keys, null-clearing, polymorphic forms), the lenient sanitizer (salvage vs wholesale-drop, passthrough, never-throws, warns), and the pipeline (sanitize-before-merge, opt-in link resolution incl. the DB-link bugfix). Existing 8 merge tests + footer tests unchanged and green.

- ✅ Full unit suite: 570 passed / 40 files
- ✅ `tsc --noEmit`: 0 errors
- ✅ ESLint clean

### DIFF
--- a/apps/flowershow/app/api/rss/[user]/[project]/route.ts
+++ b/apps/flowershow/app/api/rss/[user]/[project]/route.ts
@@ -76,7 +76,9 @@ export async function GET(
     // missing or invalid config.json — fall back to DB config only
   }
 
-  const siteConfig = resolveSiteConfig(dbConfig, fileConfig);
+  const siteConfig = resolveSiteConfig(dbConfig, fileConfig, {
+    siteId: site.id,
+  });
 
   if (!siteConfig.enableRss) {
     return new Response('Not found', { status: 404 });
--- a/apps/flowershow/lib/site-config-schema.test.ts
+++ b/apps/flowershow/lib/site-config-schema.test.ts
@@ -1,0 +1,330 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { sanitizeSiteConfig, validateConfigPatch } from './site-config-schema';
+
+describe('validateConfigPatch — accepts valid patches', () => {
+  it('accepts an empty patch', () => {
+    expect(() => validateConfigPatch({})).not.toThrow();
+  });
+
+  it('accepts well-formed scalar fields', () => {
+    expect(() =>
+      validateConfigPatch({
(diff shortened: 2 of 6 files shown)


rufuspollock:
@olayway can we merge?

PR #1383, opened 2026-09-18
feat: built-in changelog rendering for changelog/ folders

rufuspollock:
## What

Flowershow now renders changelogs out of the box. A `changelog/` folder of dated entries (or any folder whose `README.md`/`index.md` has `layout: changelog`) becomes a full-entry timeline at the folder URL, and each entry gets its own page. No components or config are needed, and it replaces the `<List>`-based approach.

- **Index** (`/changelog`): full entries, newest first, with a sticky date/authors column, title, description, optional hero image and full body. 10 per page via `?page=N`, and an out-of-range page returns 404. The folder README becomes the intro. A folder with no README still gets a "Changelog" page.
- **Entry page** (`/changelog/<slug>`): a "← Changelog" back link, the entry, and previous/next links.
- **Authors** resolve through `people/` exactly like blog posts (full name, avatar, link), with one batched `getAuthors` call per page.
- **Opt-out:** any other explicit `layout` on the folder README, or on an entry itself.
- **Theming:** stable `.changelog-*` classes, styled in `default-theme.css` with existing tokens, and documented in the generated theme class reference (the generator gets a new "Changelog" group).
- **Dogfood:** flowershow.app's own `/changelog` now uses it (`changelog/README.mdx` → `README.md`, `<List>` removed).

Single-file `CHANGELOG.md` support is phase 2 and is being planned separately.

## How

Everything happens at render time in the Next app. There are no worker, sync or schema changes.

- `lib/changelog.ts`: pure helpers (detection, date/title fallbacks, sorting, pagination, neighbours).
- `lib/changelog-context.ts`: decides index / entry / normal page for a request.
- `site.getChangelogEntries` tRPC procedure (cached, `assertSiteAccess`, Prisma `findMany`, no raw SQL).
- `lib/render-page-content.tsx`: **refactor** that extracts the markdown/MDX/canvas compile step from the site page route, so changelog entries compile through exactly the same pipeline. Worth a careful look: it touches every page render. Behaviour is unchanged, and staging's canvas `containerHeight` change is carried over.
- `components/public/changelog/*`: server components.

## Docs and design

- Design spec: `docs/plans/2026-09-18-changelog-rendering-design.md`
- Research (Linear, Mintlify, Raycast, Vercel, Keep a Changelog…): `docs/plans/2026-09-18-changelog-rendering-research.md`
- Implementation plan: `docs/plans/2026-09-18-changelog-rendering-implementation.md`
- User docs: `content/flowershow-app/docs/reference/changelog.md`
[...]

PR #1388, opened 2026-09-19
feat(changelog): render single-file CHANGELOG.md as a timeline (phase 2)

rufuspollock:
## What

Phase 2 of changelog rendering, building on #1383 (merged). It includes the two phase-2 plan commits that were pushed to `feat/changelog-rendering` after #1383 merged. A single `CHANGELOG.md` now renders as the same changelog timeline as a `changelog/` folder, on one page with an anchor per version.

- **Detection:** `changelog.md`/`.mdx` in any case, anywhere, renders as a timeline. Any other page opts in with `layout: changelog`, and a different explicit `layout` opts out. Pages inside a changelog folder are always that folder's entries; the folder decides, as in #1386. Other file names (`HISTORY.md`, `NEWS.md`…) are never guessed. A `CHANGELOG.md` with no version headings stays a normal page.
- **Formats:** every `##` starts an entry. This covers Keep a Changelog (`## [1.2.0] - 2026-09-18`), Changesets (`## 2.3.0` under `# @scope/pkg`), release-please/conventional (`## [17.11.2](…) (2026-08-24)`, plus older `#` versions), date-only (`## 2026-09-18 - Big launch`) and free-form titles. `###` subsections stay in the entry body.
- **Output:** the same DOM and classes as v1, so themes and the `--changelog-*` custom properties apply unchanged. The first `#` is the page title, and anything before the first version becomes the intro. Unreleased is shown muted (`.is-unreleased`), and skipped when it's empty.
- **URLs:** `CHANGELOG.md` is served at `/CHANGELOG` and also at `/changelog` (same for subfolders). A `changelog/` folder with entries wins `/changelog`, with or without a README.
- **Compare link:** a version heading that links somewhere (release-please inline links, Keep a Changelog reference links) gets a small "Compare" link in the meta column (`.changelog-entry-compare`). The label is "Release" if the URL isn't a `/compare/` URL.
- **Anchors:** clean ids such as `#1.2.0`, `#unreleased`, `#2026-09-18` or a title slug, deduped with `-1`, `-2`.
- **Docs:** there's a new "Single `CHANGELOG.md` file" section in `docs/reference/changelog.md`, a changelog entry, and an updated theme class reference.

## How

It all happens at render time. There are no worker, sync or schema changes.

- `lib/changelog-file.ts`: a pure heading parser (`parseVersionHeading`, `splitChangelogTree`, `entryAnchor`, `hasVersionSections`).
- `lib/remark-changelog.ts`: a remark plugin that rewrites the whole mdast once into the v1 structure using `data.hName`/`hProperties` mdast nodes, never raw HTML from user text. Because it's one compile, reference-link definitions at the bottom of Keep a Changelog files keep working. It's registered last among the remark plugins in both `processMarkdown` and `getMdxOptions`, only when the `changelog` option is set.
- `lib/changelog-context.ts`: a new `kind: 'file'`. The page route renders it in a plain `div#mdxpage`, with no `BlogLayout`, TOC or hero.
- `site.getBlob`: gains the `/changelog` alias and the folder-wins rule. **Worth a careful look:** this is the shared page lookup. The change is limited to slugs whose last segment is `changelog` (any case), apart from one general tweak. When several blobs share an `appPath`, a `README.*` now beats a plain file, after `index.*` as before. This only matters when `foo.md` and `foo/README.md` coexist.
- `rehype-autolink-headings` now skips the generated `.changelog-title` and `.changelog-entry-title`, which already link to their anchor. Entry titles get an explicit `<anchor>-title` id, so `rehype-slug` can't duplicate the `li` id.

## Decisions made during implementation (unattended run)

Rufus locked the design answers beforehand (see the "Decisions" section of `docs/plans/2026-09-19-changelog-single-file-design.md`). These smaller calls were made during the run:

1. **Folder-wins is case-sensitive, matching URLs.** `CHANGELOG.md` next to a `changelog/` folder isn't a collision: the file stays at `/CHANGELOG` and the folder owns `/changelog`. The alias applies only to lowercase `/changelog`.
2. **"Compare" vs "Release" label:** heading links without `/compare/` in the URL (e.g. release tag pages) are labelled "Release" rather than mislabelled "Compare".
3. **No-version fallback:** the design said such files render normally. The route checks this with `hasVersionSections` (one extra remark parse, after stripping frontmatter) so the page keeps `BlogLayout` and prose styles.
4. **Entry-title ids** are `<anchor>-title` (see How).
5. **#1383 had already merged** (09:56 CEST) when this run started, and #1386 landed after. So this PR is rebased onto `staging` and targets it, rather than being stacked on `feat/changelog-rendering` as planned. It's aligned with #1386: single-file detection only applies outside changelog folders, so `changelog/x.md` with `layout: changelog` in an opted-out folder stays a normal page, as #1386's test expects.

## Verification

- **Unit:** 713 of 713 pass locally on top of `staging`. They cover the parser against real heading formats, plugin HTML output, reference-link resolution, compare links, and both the MD and MDX pipelines. A **parity test** checks the plugin's DOM skeleton against the React `ChangelogEntry` and `ChangelogIndex`. The `getBlob` alias and folder-precedence cases are covered in the router tests.
- `tsc` (app and e2e), ESLint (0 errors), `docs:theme-classes:check` and the class generator tests are all clean.
- **E2E:** a new `e2e/specs/changelog-file.spec.ts` with fixtures (a root `CHANGELOG.md`, which sits next to the existing `changelog/` folder and so also tests folder-wins; `packages/cli/CHANGELOG.md`, Changesets style, checked at both URLs; and a `notes/changelog.md` opt-out). **CI:** Lint ✅ and test ✅. On E2E, all 4 new `changelog-file` tests pass in CI, including the `/packages/cli/changelog` alias, folder-wins at `/changelog` and the opt-out. The existing folder `changelog` spec also passes. The E2E job is still red because of 2 specs that fail identically on `staging`'s own latest run (`b1fbed2f`, before this PR): `custom-head.spec` (the custom head is now Premium-gated, so the test site no longer gets it) and `links-and-embeds` Links (it expects `/_next/image?url=`, but non-prod deployments now skip image optimization). Both were introduced on `staging` and are unrelated to changelogs; they need their own fix.
- **Vercel:** the `flowershow-cloud-new` preview build passed. `flowershow-components` fails here too, but it also fails on #1383, so that failure predates this PR.
- **Visual:** real Keep a Changelog, Changesets and release-please files were rendered through `processMarkdown` with the compiled default CSS and screenshotted headless at 1280px and 390px, in light and dark. The layout matches v1 and nothing overflows.
- **Q3 page weight** (real release-please `CHANGELOG.md`: 3,269 lines, 259 versions): 726 KB HTML, 61 KB gzipped, about 380 ms to compile. The same file as a plain page is 770 KB and about 320 ms. The timeline adds essentially nothing, so no cap is needed. Changesets' CLI changelog (2,261 lines, 122 versions) is 435 KB, 51 KB gzipped.

## Not done yet

- A staging check on a real deployment (private `fl` publish of a folder with a copy of `apps/cli/CHANGELOG.md`). Plan Task 7 Step 4.
- Follow-up beads: a config option for pointing at changelog files or directories, and tidying Changesets noise (someday).
- The folder changelog from #1383 has no user-facing changelog entry. This PR adds one for single files only.

Draft: don't merge until CI (including E2E) is green and a staging check has been done.


rufuspollock:
@copilot resolve the merge conflicts in this pull request

